PR #2, opened 2018-06-08
Update for Doctrine Coding Standards 4.0

jwage:


composer.lock
```diff
@@ -2567,18 +2568,57 @@
...
+            "license": [
+                "MIT"
+            ],
+            "description": "Slevomat Coding Standard for PHP_CodeSniffer complements Consistence Coding Standard by providing sniffs with additional checks.",
+            "time": "2018-05-15T21:21:12+00:00"
+        },
         {
             "name": "squizlabs/php_codesniffer",
-            "version": "3.2.3",
+            "version": "3.3.0",
```
Majkl578: Can you keep the `3.2.3` version here the same way as we did last time?

lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
```diff
@@ -1,56 +1,42 @@
...
 namespace Doctrine\Common\Persistence;
 
+use function explode;
+use function sprintf;
+use function strpos;
+
 /**
  * Abstract implementation of the ManagerRegistry contract.
  *
  * @link   www.doctrine-project.org
```
Majkl578: We can safely remove all `@link   www.doctrine-project.org`.
(This annotation isn't forbidden by our CS as it may contain useful information so it needs to be done manually - one nice regexp :) .)

lib/Doctrine/Common/Persistence/Event/OnClearEventArgs.php
```diff
@@ -8,24 +8,17 @@
...
-     * @var string|null
-     */
+    /** @var string|null */
     private $entityClass;
 
     /**
-     * Constructor.
+     *
      *
      * @param ObjectManager $objectManager The object manager.
```
Majkl578: Unfortunately we currently have no sniff for phpDoc spacing so some fixers leave superfluous empty lines:
- empty line(s) after `/**`
- empty line(s) before `*/
- multiple empty lines instead of single empty line

Can you grep through these manually? It's slightly annoying, but doable with three regexes.

PR #6, opened 2018-06-13
Remove doctrine/common backref and use specific packages instead.

jwage:
Update `doctrine/reflection` from `1.0@dev` to `^1.0` once the v1.0.0 tag is created.

### DIFF
--- a/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -4,12 +4,14 @@ namespace Doctrine\Common\Persistence\Mapping;
 
 use Doctrine\Common\Cache\Cache;
 use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
-use Doctrine\Common\Util\ClassUtils;
+use Doctrine\Common\Persistence\Proxy;
 use ReflectionException;
 use function array_reverse;
 use function array_unshift;
 use function explode;
 use function strpos;
+use function strrpos;
+use function substr;
 
 /**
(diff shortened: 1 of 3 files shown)


Majkl578:
Looks good, let's just wait for Reflection 1.0 tag.

Any specific reason why earlier Collections/Annotations/Cache are not allowed?

jwage:
@Majkl578 I can change that. What version is the min I should require? I added them to composer.json with composer require so it just added the latest version.

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -394,4 +396,22 @@ public function getReflectionService()
...
     }
+
+    /**
+     * Gets the real class name of a class name that could be a proxy.
+     *
+     * @param string $class
+     *
+     * @return string
+     */
+    private function getRealClass($class)
```
alcaeus: Shouldn't this be part of a `doctrine/proxy` implementation?

Majkl578: We don't really want doctrine/proxy, it's to be dropped soon.

alcaeus: Gotcha 👍

Majkl578: We should make this look like:
```php
private static function getRealClass(string $class) : string
```

jwage: @Majkl578 done

PR #12, opened 2018-06-26
Fix: No need to update composer itself twice

localheinz:
This PR

* [x] removes a redundant run of `composer self-update` on Travis

💁‍♂️ For reference, see

* https://travis-ci.org/doctrine/persistence/jobs/392427784#L480-L482
* https://travis-ci.org/doctrine/persistence/jobs/392427784#L510-L511


localheinz:
Thank you, @jwage!

PR #17, opened 2018-09-25
Apply Doctrine CS 5.0

Majkl578:


lib/Doctrine/Common/Persistence/Mapping/Driver/StaticPHPDriver.php
```diff
@@ -62,6 +65,7 @@ public function loadMetadataForClass($className, ClassMetadata $metadata)
 
     /**
      * {@inheritDoc}
```
alcaeus: Wasn’t there something about inheritdoc being wrong in general?

Majkl578: Yes, but we don't have that rule. It's a problem for other PHPCS rules that we use, since PHPCS has only single file scope. We'll need to find some solution/workaround (cc @kukulich) and hopefully get rid of inheritDocs later.

PR #18, opened 2018-10-13
Add NotifyPropertyChanged, PropertyChangedListener

olvlvl:
Fix https://github.com/doctrine/common/issues/855

### DIFF
--- a/lib/Doctrine/Common/NotifyPropertyChanged.php
+++ b/lib/Doctrine/Common/NotifyPropertyChanged.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Doctrine\Common;
+
+/**
+ * Interface for classes that notify event listeners of changes to their managed properties.
+ *
+ * This interface is implemented by objects that manually want to notify their object manager or
+ * other listeners when properties change, instead of relying on the object manager to compute
+ * property changes itself when changes are to be persisted.
+ */
+interface NotifyPropertyChanged
+{
+    /**
+     * Adds a listener that wants to be notified about property changes.
+     *
+     * @return void
+     */
+    public function addPropertyChangedListener(PropertyChangedListener $listener);
+}
--- a/lib/Doctrine/Common/PropertyChangedListener.php
+++ b/lib/Doctrine/Common/PropertyChangedListener.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Doctrine\Common;
+
+/**
+ * Contract for classes that are potential listeners of a {@see NotifyPropertyChanged}
+ * implementor.
+ */
+interface PropertyChangedListener
+{
+    /**
+     * Collect information about a property change.
+     *
+     * @param object $sender       The object on which the property changed.
+     * @param string $propertyName The name of the property that changed.
+     * @param mixed  $oldValue     The old value of the property that changed.
+     * @param mixed  $newValue     The new value of the property that changed.
+     *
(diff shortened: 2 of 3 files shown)


Majkl578:
Hi,
thanks for working on this! I restarted the crashed build, can you please see & address the [CS issues](https://travis-ci.org/doctrine/persistence/jobs/440956588)? Thanks.

Majkl578:
Also please bump the `conflict` here:
https://github.com/doctrine/persistence/blob/b8b8b9398323cef06cc7def32a0f217e6ec7b587/composer.json#L36
to `<2.10@dev`.

olvlvl:
@Majkl578 This time, I ran `./vendor/bin/phpcs` before committing :)

lib/Doctrine/Common/NotifyPropertyChanged.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Doctrine\Common;
+
+/**
+ * Contract for classes that provide the service of notifying listeners of
```
Ocramius: This description needs improvement: currently, you are replicating the name of the class in it. 

Also try to read it by assuming no previous knowledge about it

lib/Doctrine/Common/PropertyChangedListener.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Doctrine\Common;
+
+/**
+ * Contract for classes that are potential listeners of a <tt>NotifyPropertyChanged</tt>
```
Ocramius: Please use `{@see MyClass}`

Ocramius: The cyclical dependency of these classes is a bit hard to grasp: can we add another sentence or two that don't involve explaining this with `NotifyPropertyChanged` in mind?

lib/Doctrine/Common/PropertyChangedListener.php
```diff
@@ -0,0 +1,22 @@
...
+ * implementor.
+ */
+interface PropertyChangedListener
+{
+    /**
+     * Notifies the listener of a property change.
```
Ocramius: "Collect information about a property change"

lib/Doctrine/Common/NotifyPropertyChanged.php
Ocramius: > What would you recommend as a description for this interface?

That's what I asked you to come up with ;-)

Majkl578:
@Ocramius I'd personally prefer doing such changes in separate PR, leaving this only a _move stuff from X to Y_.

olvlvl:
I second that :)

Ocramius:
If it's got to be committed, and it counts as an addition, let's use the chance to improve that.

We all know very well that otherwise it just lands in the endless wastelands of "I wish we did also X".

lib/Doctrine/Common/NotifyPropertyChanged.php
alcaeus: Suggestion:

> Interface for classes that notify event listeners of changes to their managed properties.
>
> This interface is implemented by objects that manually want to notify their object manager or other listeners when properties changes, instead of relying on the object manager to compute property changes itself when changes are to be persisted.

olvlvl:
I updated the dock-block with @alcaeus suggestion.

PR #21, opened 2018-10-20
Fix Metadata Caching when it changes in EventListeners

andrey-bondar-dron:
This PR fixed 2 bugs:
1. Using https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/cookbook/resolve-target-entity-listener.html with metadata cache is not working, because ResolveTargetEntityListener calls setMetadataFor without any effect on generated cache.
2. Fallback metadata is not caching now.

I hope, it will be merged to version 1.0.2 and released ASAP.


tests/Doctrine/Tests/Common/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -159,6 +159,35 @@ public function testWillIgnoreCacheEntriesThatAreNotMetadataInstances()
 
         self::assertSame($metadata, $this->cmf->getMetadataFor('Foo'));
     }
+
+    public function testFallbackMetadataShouldBeCached()
+    {
+        /** @var Cache|PHPUnit_Framework_MockObject_MockObject $cacheDriver */
```
lcobucci: We should rather rely on `PHPUnit\Framework\MockObject\MockObject` since it's part of PHPUnit's forward compatibility layer (even though the rest of the test uses the old interface).

tests/Doctrine/Tests/Common/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -159,6 +159,35 @@ public function testWillIgnoreCacheEntriesThatAreNotMetadataInstances()
...
+        $cacheDriver = $this->createMock(Cache::class);
+        $cacheDriver->expects($this->once())->method('save');
```
lcobucci: `once()` is a static method, so we shouldn't call it dynamically.

```suggestion
        $cacheDriver->expects(self::once())->method('save');
```

tests/Doctrine/Tests/Common/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -159,6 +159,35 @@ public function testWillIgnoreCacheEntriesThatAreNotMetadataInstances()
...
+        $this->cmf->fallbackCallback = static function () use ($classMetadata) {
+            return $classMetadata;
+        };
+
+        $this->cmf->getMetadataFor('Foo');
+    }
+
+    public function testSetMetadataForShouldUpdateCache()
+    {
+        /** @var Cache|PHPUnit_Framework_MockObject_MockObject $cacheDriver */
```
lcobucci: We should rather rely on `PHPUnit\Framework\MockObject\MockObject` since it's part of PHPUnit's forward compatibility layer (even though the rest of the test uses the old interface).

tests/Doctrine/Tests/Common/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -159,6 +159,35 @@ public function testWillIgnoreCacheEntriesThatAreNotMetadataInstances()
...
+        $cacheDriver = $this->createMock(Cache::class);
+        $cacheDriver->expects($this->once())->method('save');
```
lcobucci: `once()` is a static method, so we shouldn't call it dynamically.

```suggestion
        $cacheDriver->expects(self::once())->method('save');
```

lcobucci:
Can you also rebase your branch to organise your commits, please?

tests/Doctrine/Tests/Common/Persistence/Mapping/ClassMetadataFactoryTest.php
malarzm: I'd rather stay with `$this->once()`, even [PHPUnit's docs](https://phpunit.readthedocs.io/en/7.3/test-doubles.html) is using this format. Also I haven't seen `self::once()` until today I think :D

lcobucci: @malarzm we're using that in other repos (and even in this class) :+1:

lcobucci: This is also a requirement for `phpstan-strict-rules` BTW :grin:

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -234,6 +230,16 @@ public function hasMetadataFor($className)
     public function setMetadataFor($className, $class)
     {
         $this->loadedMetadata[$className] = $class;
+
+        if (! $this->cacheDriver) {
```
Majkl578:
```suggestion
        if ($this->cacheDriver === null) {
```
Handling this in a rebase too.

andrey-bondar-dron:
Hi! Guys, someone has already fixed all required changes, while I was absent. Thanks. :)

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -167,7 +167,9 @@ public function getMetadataFor($className)
 
         if (isset($this->loadedMetadata[$realClassName])) {
             // We do not have the alias name in the map, include it
-            return $this->loadedMetadata[$className] = $this->loadedMetadata[$realClassName];
+            $this->setMetadataFor($realClassName, $this->loadedMetadata[$realClassName]);
```
Majkl578: This particular change breaks in-memory caching (`$this->loadedMetadata`) for aliased classes, therefore breaking hot path at the top of this method for those:
```php
public function testGettingAliasesOnlyFetchesFromCacheOnce() : void
{
    $loadedReflection = new ReflectionProperty(AbstractClassMetadataFactory::class, 'loadedMetadata');
    $loadedReflection->setAccessible(true);

    $this->cmf->getMetadataFor(RootEntity::class);
    self::assertArrayHasKey(RootEntity::class, $loadedReflection->getValue($this->cmf));
    self::assertArrayNotHasKey('Alias:RootEntity', $loadedReflection->getValue($this->cmf));

    $this->cmf->getMetadataFor('Alias:RootEntity');
    self::assertArrayHasKey(RootEntity::class, $loadedReflection->getValue($this->cmf));
    self::assertArrayHasKey('Alias:RootEntity', $loadedReflection->getValue($this->cmf));
}
```
It's because this condition has changed semantics and no longer does what the comment above says: "_We do not have the alias name in the map, include it_"

andrey-bondar-dron: @Majkl578 When it is needed to change metadata for some entity in EventListeners, we do not know about aliases that were used.
When I call setMetadataFor with real class name, AbstractClassMetadataFactory should replace metadata for all aliases in $loadedMetadata. And vise versa when I call setMetadataFor with alias.
For my opinion, it is normal when we will save only real classes in $loadedMetadata, because it prevent to have many problems with aliases.
On every call getMetadataFor with alias this code will be executed
```php
// Check for namespace alias
        if (strpos($className, ':') !== false) {
            [$namespaceAlias, $simpleClassName] = explode(':', $className, 2);

            $realClassName = $this->getFqcnFromAlias($namespaceAlias, $simpleClassName);
        } else {
            $realClassName = $this->getRealClass($className);
        }
```
In order to prevent it I can create map for 'alias' => 'realClassName' and replace this code
```php
        // Check for namespace alias
        if (strpos($className, ':') !== false) {
            [$namespaceAlias, $simpleClassName] = explode(':', $className, 2);

            $realClassName = $this->getFqcnFromAlias($namespaceAlias, $simpleClassName);
        } else {
            $realClassName = $this->getRealClass($className);
        }
```
with something like this
```php
        switch (true) {
            case isset($this->aliasesMap[$className]):
                $realClassName = $this->aliasesMap[$className];
                break;
            case strpos($className, ':') !== false // Check for namespace alias
                [$namespaceAlias, $simpleClassName] = explode(':', $className, 2);
                $realClassName = $this->getFqcnFromAlias($namespaceAlias, $simpleClassName);
                break;
            default:
                $realClassName = $this->getRealClass($className);
        }
```
In this case AbstractClassMetadataFactory will not be too coupled with aliases in setMetadataFor/getMetadataFor methods.

andrey-bondar-dron: So, fixed.
For now in in-memory cache (also in MetadataCache) will be saved single value for all aliases. 
Calling setMetadataFor and getMetadataFor with aliases or FQCN will same keys.

andrey-bondar-dron:
@Majkl578 @lcobucci @alcaeus @malarzm Hi all! Give please your feedback.

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -399,16 +391,32 @@ public function getReflectionService()
...
     /**
-     * Gets the real class name of a class name that could be a proxy.
+     * Gets the real class name of a class name that could be a proxy or alias.
+     *
+     * @param string $className
+     *
+     * @return string
      */
-    private function getRealClass(string $class) : string
+    private function getRealClassName($className)
```
alcaeus: Type hints are here to stay: there's no reason you should assume anything else than a string for `$className`.

alcaeus: We may also profit from making this method `protected`: class name resolution for proxy class names is currently more difficult in ODM 2.0 and ORM 3.0 (master) since moving away from the legacy proxies. Overriding this method would at least temporarily solve this problem. Opinions @lcobucci @Majkl578?

PR #35, opened 2019-01-02
Open 2.x

Majkl578:
* Require PHP 7.2.
* Remove "Common" from namespace.
* Move NotifyPropertyChanged and PropertyChangedListener into Persistence namespace.


Ocramius:
> These should be simplified to `lib` and `tests` respectively.

This is intentionally left as such due to Doctrine's PSR-0 directory tree. While PSR-4 is being used (we don't use `_` anymore), the prefixing is precisely the reason why @doctrine voted against PSR-4.

PR #41, opened 2019-02-18
Metadata cache lifetime should be int

dron-alterpost:
\Doctrine\Common\Cache\Cache::save does not accept null, but AbstractClassMetadataFactory does it.
The problem occurs with usage of \Doctrine\Common\Cache\PhpFileCache in 
```php
private function includeFileForId(string $id) : ?array
    {
        $fileName = $this->getFilename($id);

        // note: error suppression is still faster than `file_exists`, `is_file` and `is_readable`
        set_error_handler(self::$emptyErrorHandler);

        $value = include $fileName;

        restore_error_handler();

        if (! isset($value['lifetime'])) {
            return null;
        }

        return $value;
    }
```

on ```! isset($value['lifetime'])```


greg0ire:
Link for the lazy: https://github.com/doctrine/cache/blob/6f9810e194acace22e360a1398cf228132e5b4b2/lib/Doctrine/Common/Cache/Cache.php#L55

PR #42, opened 2019-02-18
Fix Metadata Caching when it changes in EventListeners

andrey-bondar-dron:
I've closed PR https://github.com/doctrine/persistence/pull/21 and created new one, because that PR is targeted on master.

This PR fixed 2 bugs:

1. Using https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/cookbook/resolve-target-entity-listener.html with metadata cache is not working, because ResolveTargetEntityListener calls setMetadataFor without any effect on generated cache.
2. Fallback metadata is not caching now.

### DIFF
--- a/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -175,10 +175,10 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
                     $this->loadedMetadata[$realClassName] = $cached;
                     $this->wakeupReflection($cached, $this->getReflectionService());
                 } else {
-                    $this->loadMetadata($className);
+                    $this->loadMetadata($realClassName);
                 }
             } else {
-                $this->loadMetadata($className);
+                $this->loadMetadata($realClassName);
             }
         } catch (MappingException $loadingException) {
             $fallbackMetadataResponse = $this->onNotFoundMetadata($className);
@@ -188,6 +188,10 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
             }
 
             $this->setMetadataFor($className, $fallbackMetadataResponse);
+        }
+
+        if ($className !== $realClassName) {
+            $this->setMetadataFor($className, $this->loadedMetadata[$realClassName]);
         }
 
         return $this->loadedMetadata[$className];
--- a/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -36,7 +36,7 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
     private $loadedMetadata = [];
 
     /** @var string[] */
-    protected $aliasesMap = [];
+    private $aliasesMap = [];
 
     /** @var bool */
     protected $initialized = false;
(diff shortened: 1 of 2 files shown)


Pavlentiycomua (dismissed):

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -217,6 +198,8 @@ public function getMetadataFor($className)
      */
     public function hasMetadataFor($className)
     {
+        $className = $this->getRealClassName($className);
```
Ocramius: Please revert this one: overhead should stay very minimal here, and indeed, duplicates in `$this->loadedMetadata` are expected

Ocramius (changes requested):
Solution looks valid, but we need to duplicate entries in `aliasesMap` to simplify "happy path" lookups, and reduce overhead there.

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -152,57 +155,35 @@ public function getAllMetadata()
      */
     public function getMetadataFor($className)
     {
+        $className = $this->getRealClassName($className);
```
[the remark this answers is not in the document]
dron-alterpost: @Ocramius Thank you for review! 
But when listener changes metadata, then metadata should be changed for all class aliases, but listener does not know about all variants of aliases.
If metadata changed for "AcmeBundle:MyEntity", then it must be changed for "AcmeBundle\Entity\MyEntity" and vice versa.

How to handle this case?

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -398,16 +390,28 @@ public function getReflectionService()
...
+    protected function getRealClassName(string $className) : string
     {
-        $pos = strrpos($class, '\\' . Proxy::MARKER . '\\');
+        if (isset($this->aliasesMap[$className])) {
+            return $this->aliasesMap[$className];
+        }
 
-        if ($pos === false) {
-            return $class;
+        switch (true) {
```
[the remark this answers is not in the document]
dron-alterpost: Done.

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -35,6 +35,9 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
     /** @var ClassMetadata[] */
     private $loadedMetadata = [];
 
+    /** @var string[] */
+    protected $aliasesMap = [];
```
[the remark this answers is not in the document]
dron-alterpost: Done.

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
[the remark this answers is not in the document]
Ocramius: Multiple hits with different names (even though logical name is the same) can either lead to metadata loading or cache loading

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -398,16 +395,28 @@ public function getReflectionService()
     }
 
     /**
-     * Gets the real class name of a class name that could be a proxy.
+     * Gets the real class name of a class name that could be a proxy or alias.
      */
-    private function getRealClass(string $class) : string
+    protected function getRealClassName(string $className) : string
```
Ocramius: This should most likely stay `private`

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -398,16 +395,28 @@ public function getReflectionService()
...
     {
-        $pos = strrpos($class, '\\' . Proxy::MARKER . '\\');
+        if (isset($this->aliasesMap[$className])) {
```
Ocramius: Considering that duplicate entries are allowed, is this property still needed?

Ocramius (changes requested):

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
ostrolucky: This is protected as per suggestion of @alcaeus at https://github.com/doctrine/persistence/pull/21#discussion_r249341149

lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
ostrolucky: This was done because of https://github.com/doctrine/persistence/pull/21#discussion_r235214384 Can you please decipher comment chain and explain if something still needs to be done?

PR #44, opened 2019-03-19
Remove dupes in FileDriver::getAllClassNames

NathanBaulch:
`FileDriver::getAllClassNames()` should only return unique class names when some classes are already cached.

### DIFF
--- a/lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
+++ b/lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
@@ -6,6 +6,7 @@ use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\MappingException;
 use function array_keys;
 use function array_merge;
+use function array_unique;
 use function is_file;
 use function str_replace;
 
@@ -126,10 +127,10 @@ abstract class FileDriver implements MappingDriver
             return (array) $this->locator->getAllClassNames($this->globalBasename);
         }
 
-        return array_merge(
+        return array_unique(array_merge(
             array_keys($this->classCache),
             (array) $this->locator->getAllClassNames($this->globalBasename)
-        );
+        ));
     }
 
     /**
--- a/tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
+++ b/tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
@@ -101,6 +101,22 @@ class FileDriverTest extends DoctrineTestCase
         self::assertSame(['stdGlobal', 'stdGlobal2', 'stdClass'], $classNames);
     }
 
+    public function testGetAllClassNamesBothSourcesNoDupes() : void
+    {
+        $locator = $this->newLocator();
+        $locator->expects($this->once())
+                ->method('getAllClassNames')
+                ->with($this->equalTo('global'))
+                ->willReturn(['stdClass']);
+        $driver = new TestFileDriver($locator);
+        $driver->setGlobalBasename('global');
+
+        $driver->getElement('stdClass');
+        $classNames = $driver->getAllClassNames();
+
+        self::assertSame(['stdGlobal', 'stdGlobal2', 'stdClass'], $classNames);
+    }
+
     public function testIsNotTransient()
     {
         $locator = $this->newLocator();


tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
```diff
@@ -101,6 +101,22 @@ public function testGetAllClassNamesBothSources()
         self::assertSame(['stdGlobal', 'stdGlobal2', 'stdClass'], $classNames);
     }
 
+    public function testGetAllClassNamesBothSourcesNoDupes()
```
Ocramius: `: void`

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
```diff
@@ -126,10 +127,10 @@ public function getAllClassNames()
             return (array) $this->locator->getAllClassNames($this->globalBasename);
         }
 
-        return array_merge(
+        return array_unique(array_merge(
```
Ocramius: Does the problem only occur with thr `FileDriver`?

tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
NathanBaulch: Sure, on just the new test function or all of them in that class?

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
NathanBaulch: Yes, I just checked and this is the only place in the project where `array_merge` is used with indexed/numeric arrays without `array_unique`.

tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
Ocramius: Just this one - we try to fix CS gradually, on modified code only, but we may improve this in one shot later on :)

Ocramius:
Commit 4ba156c  should not be in this patch - can you drop it with an interactive rebase?

tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
NathanBaulch: Done.

NathanBaulch:
> Commit [4ba156c](https://github.com/doctrine/persistence/commit/4ba156c2af86895a68f1b357f58eb6664caff574) should not be in this patch - can you drop it with an interactive rebase?

You lost me, that's an almost 8 year old commit! It's only significance is that's where the original problem was introduced, so that's where I first commented on it.

Ocramius:
Oh, sorry, I was misguided by the Github Mobile UI there...

NathanBaulch:
That was going to be an epic interactive rebase! :P

NathanBaulch:
Need anything else from me?

tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
```diff
@@ -101,6 +101,22 @@ public function testGetAllClassNamesBothSources()
...
+    {
+        $locator = $this->newLocator();
+        $locator->expects($this->any())
```
SenseException:
```suggestion
        $locator->expects($this->once())
```

tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
```diff
@@ -101,6 +101,22 @@ public function testGetAllClassNamesBothSources()
...
+                ->method('getAllClassNames')
+                ->with($this->equalTo('global'))
+                ->will($this->returnValue(['stdClass']));
```
SenseException:
```suggestion
                ->willReturn(['stdClass']);
```

tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
SenseException: Could it be that you added `void` to a totally different test method instead of your own?

NathanBaulch:
Bump!

tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
NathanBaulch: Not sure why I did that, fixed!

ostrolucky:
This should target 1.1.x Can you please rebase?

ostrolucky:
Did the rebase on my own in https://github.com/doctrine/persistence/pull/56

PR #47, opened 2019-04-10
Deprecate ObjectManager::merge() and ObjectManager::detach()

Majkl578:
Added deprecation for `ObjectManager::merge()` and `ObjectManager::detach()` ahead of their removal in 2.0 and ORM 3.0.

More context in:
* https://github.com/doctrine/orm/blob/ebb53acda96e304a6f91d4a5a2d76378a4e1d210/UPGRADE.md#bc-break-removed-entitymanagermerge-and-entitymanagerdetach-methods
* https://github.com/doctrine/orm/pull/1577

### DIFF
--- a/lib/Doctrine/Common/Persistence/ObjectManager.php
+++ b/lib/Doctrine/Common/Persistence/ObjectManager.php
@@ -52,6 +52,9 @@ interface ObjectManager
      * of this ObjectManager and returns the managed copy of the object.
      * The object passed to merge will not become associated/managed with this ObjectManager.
      *
+     * @deprecated Merge operation is deprecated and will be removed in Persistence 2.0. Merging should be part of
+     *             the business domain of an application rather than a generic operation of ObjectManager.
+     *
      * @param object $object
      *
      * @return object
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Persistence/ObjectManager.php
```diff
@@ -52,6 +52,9 @@ public function remove($object);
      * of this ObjectManager and returns the managed copy of the object.
      * The object passed to merge will not become associated/managed with this ObjectManager.
      *
+     * @deprecated Merge operation is deprecated and will be removed in Persistence 2.0. Merging should be part of
+     *             business domain of an application rather than a generic operation of ObjectManager.
```
greg0ire:
```suggestion
     *             the business domain of an application rather than a generic operation of ObjectManager.
```

alcaeus:
Before merging this, should we add `UPGRADE` files to document all deprecations/removals? Symfony does it this way (see relevant files for [Symfony 4.3](https://github.com/symfony/symfony/blob/master/UPGRADE-4.3.md) and [Symfony 5.0](https://github.com/symfony/symfony/blob/master/UPGRADE-5.0.md)) and it gives users a quick way of seeing what might be broken even before installing a new version. Asking because this is a manual process that maintainers need to keep in mind when reviewing pull requests.

Majkl578:
We use single UPGRADE.md document in other repos, I'd stick with that.

malarzm:
> We use single UPGRADE.md document in other repos, I'd stick with that.

Not everywhere, in ODM we have an UPDATE file per version: https://github.com/doctrine/mongodb-odm/tree/1.3.x

PR #48, opened 2019-04-15
Use getManagerForClass in getRepository

### LINKED ISSUE #45: `ManagerRegistry#getRepository()` does not use `getManagerForClass()` internally
Assuming the `Foo` class is not in the default entity manager:

- works:

```php
$objectManager = $managerRegistry->getManagerForClass(Foo::class);
$repository = $objectManager->getRepository(Foo::class);
```

- works:

```php
$repository = $managerRegistry->getRepository(Foo::class, <manager name>);
```

- doesn't work (because `Foo` is looked into the default manager):

```php
$repository = $managerRegistry->getRepository(Foo::class);
```

I would have thought the 3rd case would work: after all the manager registry can retrieve the correct manager using `getManagerForClass()` internally.

---

Would it make sense to have `getRepository()` use `getManagerForClass()` internally to retrieve the repository in the correct manager?

alcaeus:
Fixes #45. Targeting 1.1.x since it's unexpected behaviour (thus considered a bug).

This changes logic in `getRepository` to use `getManagerForClass` if no repository was specified. Basically, instead of just using a specific manager (falling back to the default if none was given), we now use the following process:
* If a `persistentManagerName` is given, we always call `getRepository` on that manager
* If no `persistentManagerName` is given, we call `getManagerForClass` to find the first object manager that manages the given object class
* If the previous step did not yield an object manager, we fall back to the default object manager.

Note: changes from the first commit were necessary to be able to run tests standalone - somehow the PHPUnit config did not play nice with that.

### DIFF
--- a/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
@@ -210,7 +210,9 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
      */
     public function getRepository($persistentObjectName, $persistentManagerName = null)
     {
-        return $this->getManager($persistentManagerName)->getRepository($persistentObjectName);
+        return $this
+            ->selectManager($persistentObjectName, $persistentManagerName)
+            ->getRepository($persistentObjectName);
     }
 
     /**
@@ -231,5 +233,14 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
         $this->resetService($this->managers[$name]);
 
         return $this->getManager($name);
+    }
+
+    private function selectManager(string $persistentObjectName, ?string $persistentManagerName = null) : ?ObjectManager
+    {
+        if ($persistentManagerName !== null) {
+            return $this->getManager($persistentManagerName);
+        }
+
+        return $this->getManagerForClass($persistentObjectName) ?? $this->getManager();
     }
 }
--- a/tests/Doctrine/Tests/Common/Persistence/Mapping/TestClassMetadataFactory.php
+++ b/tests/Doctrine/Tests/Common/Persistence/Mapping/TestClassMetadataFactory.php
@@ -1,0 +1,81 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Common\Persistence\Mapping;
+
+use Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory;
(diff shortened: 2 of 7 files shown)


lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
```diff
@@ -210,7 +210,17 @@ public function getManagers()
      */
     public function getRepository($persistentObjectName, $persistentManagerName = null)
     {
-        return $this->getManager($persistentManagerName)->getRepository($persistentObjectName);
+        if ($persistentManagerName !== null) {
+            $manager = $this->getManager($persistentManagerName);
+        } else {
```
Majkl578: Could we avoid `else` branch and nested  `if` inside? Possibly using private method and early-exit approach.

tests/Doctrine/Tests/Common/Persistence/Mapping/TestClassMetadataFactory.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+declare(strict_types = 1);
```
Majkl578: Not following our CS.

tests/Doctrine/Tests/Common/Persistence/Mapping/TestClassMetadataFactory.php
```diff
@@ -0,0 +1,81 @@
...
+    /** @var MappingDriver */
+    public $driver;
+
+    /** @var ClassMetadata|null */
+    public $metadata;
+
+    /** @var callable|null */
+    public $fallbackCallback;
+
+    public function __construct($driver, $metadata)
```
Majkl578: Could we add typehints here?

tests/Doctrine/Tests/Common/Persistence/OtherTestObject.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types = 1);
```
Majkl578: Not following our CS.

tests/Doctrine/Tests/Common/Persistence/TestObject.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types = 1);
```
Majkl578: Not following our CS.

alcaeus:
Feedback by @Majkl578 implemented 👍

PR #50, opened 2019-04-17
Add types to all interfaces.

jwage:
- Upgrades PHPUnit to 8
- Upgrades PHPStan to 0.11
- Adds phpstan/phpstan-strict-rules
- Adds type hints everywhere and enables strict typing


ostrolucky:
persistence 2.0, orm 3.0. Then we can continue with https://github.com/doctrine/orm/pull/7130

lib/Doctrine/Persistence/Mapping/Driver/DefaultFileLocator.php
```diff
@@ -111,31 +109,37 @@ public function findMappingFile($className)
...
-                    // NOTE: All files found here means classes are not transient!
-                    $classes[] = str_replace('.', '\\', $fileName);
+                if ($fileName === $file->getBasename() || $fileName === $globalBasename) {
+                    continue;
                 }
+
+                // NOTE: All files found here means classes are not transient!
+
+                $class = str_replace('.', '\\', $fileName);
+                assert(is_string($class));
```
muglug: Psalm thinks this is always the case

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
```diff
@@ -32,11 +34,10 @@ abstract class FileDriver implements MappingDriver
...
      * documents and operates in the specified operating mode.
      *
-     * @param string|string[]|FileLocator $locator       A FileLocator or one/multiple paths
-     *                                                where mapping documents can be found.
-     * @param string|null                 $fileExtension
+     * @param string|string[]|FileLocator $locator A FileLocator or one/multiple paths
+     *                                          where mapping documents can be found.
      */
-    public function __construct($locator, $fileExtension = null)
+    public function __construct($locator, ?string $fileExtension = null)
```
muglug: I don't think other classes (e.g. `PHPDriver`) constructors _should_ automatically inherit these annotations, as constructor calls aren't subservient to the Liskov Substitution Principle. So I think these annotations need to be on every subclass with its own constructor.

jwage: I will fixup the constructors. I see what you mean now that I look at it.

lib/Doctrine/Persistence/Mapping/Driver/SymfonyFileLocator.php
```diff
@@ -55,30 +59,30 @@ class SymfonyFileLocator implements FileLocator
...
-     * @param string      $nsSeparator   String which would be used when converting FQCN to filename and vice versa. Should not be empty
+     * @param string[] $prefixes
+     * @param string   $nsSeparator String which would be used when converting FQCN to filename and vice versa. Should not be empty
      */
-    public function __construct(array $prefixes, $fileExtension = null, $nsSeparator = '.')
-    {
+    public function __construct(
+        array $prefixes,
+        ?string $fileExtension = null,
+        ?string $nsSeparator = '.'
```
ostrolucky: Any reason why we made this nullable?

lib/Doctrine/Persistence/Mapping/Driver/FileLocator.php
```diff
@@ -26,28 +24,22 @@ public function findMappingFile($className);
      *
      * @return string[]
      */
-    public function getAllClassNames($globalBasename);
+    public function getAllClassNames(?string $globalBasename = null) : array;
```
ostrolucky: Why making it nullable and having default value? It breaks contract

tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
```diff
@@ -1,164 +1,202 @@
...
     {
-        $driver = new TestFileDriver($this->newLocator());
+        $driver = $this->createTestFileDriver($this->newLocator());
+
         $driver->setGlobalBasename('global');
 
         $element = $driver->getElement('stdGlobal');
 
-        self::assertSame('stdGlobal', $element);
+        self::assertSame($driver->stdGlobal, $element);
```
ostrolucky: Can you explain this change pls? This change made this test case a bit confusing to me, as it's more dynamic and assertion depends on driver behaviour.

.travis.yml
```diff
@@ -51,4 +51,4 @@ jobs:
     - stage: Code Quality
       env: STATIC_ANALYSIS
       install: travis_retry composer install --prefer-dist
-      script: vendor/bin/phpstan analyse -l 3 -c phpstan.neon lib tests
+      script: vendor/bin/phpstan a
```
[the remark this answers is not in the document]
jwage: Yes. This is just how it is in other Doctrine libs so I made it consistent. And the other information like level and paths are in phpstan.neon.dist now to match other libs as well.

lib/Doctrine/Persistence/AbstractManagerRegistry.php
```diff
@@ -139,86 +142,107 @@ public function getDefaultManagerName()
...
      */
-    public function getManagerForClass($class)
+    public function getManagerForClass(string $class) : ?ObjectManager
     {
         // Check for namespace alias
         if (strpos($class, ':') !== false) {
             [$namespaceAlias, $simpleClassName] = explode(':', $class, 2);
-            $class                              = $this->getAliasNamespace($namespaceAlias) . '\\' . $simpleClassName;
+
+            $class = $this->getAliasNamespace($namespaceAlias) . '\\' . $simpleClassName;
```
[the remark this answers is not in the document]
jwage: Easier to read imo and having the equal sign of those two non-similar lines grouped didn't make sense. It also made the horizontal width really wide and required scrolling left and right.

lib/Doctrine/Persistence/AbstractManagerRegistry.php
```diff
@@ -139,86 +142,107 @@ public function getDefaultManagerName()
...
-        return $this
-            ->selectManager($persistentObjectName, $persistentManagerName)
-            ->getRepository($persistentObjectName);
+    public function getRepository(
+        string $persistentObjectName,
+        ?string $persistentManagerName = null
+    ) : ObjectRepository {
+        $manager = $this->selectManager($persistentObjectName, $persistentManagerName);
+
+        if ($manager === null) {
```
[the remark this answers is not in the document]
jwage: I will make this change.

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -224,30 +214,26 @@ public function hasMetadataFor($className)
...
      * @return string[]
      */
-    protected function getParentClasses($name)
+    protected function getParentClasses(string $name) : array
     {
         // Collect parent classes, ignoring transient (not-mapped) classes.
         $parentClasses = [];
 
-        foreach (array_reverse($this->getReflectionService()->getParentClasses($name)) as $parentClass) {
+        $parentClasses = $this->getReflectionService()
```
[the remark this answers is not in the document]
jwage: I got a little overzealous.

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -76,24 +81,27 @@ abstract class AnnotationDriver implements MappingDriver
...
     {
         $this->reader = $reader;
-        if (! $paths) {
+
+        if ($paths === '' || $paths === [] || $paths === null) {
             return;
         }
 
-        $this->addPaths((array) $paths);
+        if (! is_array($paths)) {
```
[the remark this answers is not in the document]
jwage: It was to get rid of the casting with `(array)`. I liked the explicit check and creation of an array. Not sure why phpstan didn't complain, but it normally will complain about casting something that is already an array, string, etc.

lib/Doctrine/Persistence/Mapping/Driver/SymfonyFileLocator.php
jwage: Hmm, let me double check that and see if I can get rid of it.

tests/Doctrine/Tests/Persistence/Mapping/FileDriverTest.php
jwage: The test before was setting up the mock to return strings on a method that should return a ClassMetadata. PHP just didn't care because of the lack of type hints. As soon as we had proper type hints, the test failed, so this is how I fixed it.

lib/Doctrine/Persistence/Mapping/Driver/FileLocator.php
jwage: We had some instances in the tests that passed null. Let me see what I can do make this better.

.travis.yml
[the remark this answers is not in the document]
stof: to be exact, this is a shortcut for a command starting with `a`, as long as there is no ambiguity. That's the behavior of symfony/console. For a CI config, using the actual name is better IMO, as this avoids breaking the CI config in case a newer version adds another command starting with `a`

tests/Doctrine/Tests/TestInit.php
```diff
@@ -16,9 +19,9 @@
 error_reporting(E_ALL | E_STRICT);
 
 // register silently failing autoloader
-spl_autoload_register(static function ($class) {
+spl_autoload_register(static function ($class) : bool {
```
stof: actually, the `return true` should be removed, and the return type should be `void`.

but this could be removed entirely by using `autoload-dev` in composer.json`

jwage: I removed this. Thanks for the suggestion!

.travis.yml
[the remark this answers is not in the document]
jwage: I changed it to be `analyze`

lib/Doctrine/Persistence/AbstractManagerRegistry.php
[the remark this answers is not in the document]
alcaeus: See #57.

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -183,7 +174,8 @@ public function getMetadataFor($className)
                     foreach ($this->loadMetadata($realClassName) as $loadedClassName) {
                         $this->cacheDriver->save(
                             $loadedClassName . $this->cacheSalt,
-                            $this->loadedMetadata[$loadedClassName]
+                            $this->loadedMetadata[$loadedClassName],
+                            0
```
alcaeus: Why are we explicitly passing a lifetime now?

lib/Doctrine/Persistence/Mapping/ClassMetadata.php
```diff
@@ -100,55 +78,37 @@ public function getIdentifierFieldNames();
...
 
     /**
      * Checks if the association is the inverse side of a bidirectional association.
-     *
-     * @param string $assocName
-     *
-     * @return bool
      */
-    public function isAssociationInverseSide($assocName);
+    public function isAssociationInverseSide(string $assocName) : bool;
```
alcaeus: Maybe refactor the parameter name to `$associationName`?
```suggestion
    public function isAssociationInverseSide(string $associationName) : bool;
```

lib/Doctrine/Persistence/Mapping/ClassMetadata.php
```diff
@@ -100,55 +78,37 @@ public function getIdentifierFieldNames();
...
 
     /**
      * Returns the target field of the owning side of the association.
-     *
-     * @param string $assocName
-     *
-     * @return string
      */
-    public function getAssociationMappedByTargetField($assocName);
+    public function getAssociationMappedByTargetField(string $assocName) : string;
```
alcaeus: Maybe refactor the parameter name to `$associationName`?
```suggestion
    public function getAssociationMappedByTargetField(string $associationName) : string;
```

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -76,24 +81,27 @@ abstract class AnnotationDriver implements MappingDriver
...
      * Appends lookup paths to metadata driver.
      *
      * @param string[] $paths
-     *
-     * @return void
      */
-    public function addPaths(array $paths)
+    public function addPaths(array $paths) : void
     {
         $this->paths = array_unique(array_merge($this->paths, $paths));
```
alcaeus: Again, we might not want to trust that `$paths` is an array of strings but rather check to make sure it is so.

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -239,9 +240,11 @@ public function getAllClassNames()
         $declared = get_declared_classes();
 
         foreach ($declared as $className) {
-            $rc         = new ReflectionClass($className);
+            $rc = new ReflectionClass($className);
+
```
alcaeus: With how often you inserted blank lines to break alignment I'd like to bring https://github.com/doctrine/coding-standard/pull/107 to your attention 😉

lib/Doctrine/Persistence/ObjectRepository.php
```diff
@@ -16,14 +18,14 @@ interface ObjectRepository
...
-    public function find($id);
+    public function find($id) : ?object;
 
     /**
      * Finds all objects in the repository.
      *
      * @return object[] The objects.
      */
-    public function findAll();
+    public function findAll() : array;
```
alcaeus: I believe we should typehint `iterable` here. For example, passing an array means hydrating every object directly after the query, while we may be able to return a lazy iterator otherwise (as is done in MongoDB ODM in some instances).
```suggestion
    public function findAll() : iterable;
```

Note: this also applies to other methods that return an `array` in this interface.

lib/Doctrine/Persistence/ObjectRepository.php
```diff
@@ -34,14 +36,17 @@ public function findAll();
...
      *
      * @throws UnexpectedValueException
      */
-    public function findBy(array $criteria, ?array $orderBy = null, $limit = null, $offset = null);
+    public function findBy(
+        array $criteria,
+        ?array $orderBy = null,
+        ?int $limit = null,
+        ?int $offset = null
+    ) : array;
```
alcaeus:
```suggestion
    ) : iterable;
```

phpstan.neon
```diff
@@ -1,7 +1,23 @@
...
         - %currentWorkingDirectory%/tests/Doctrine/Tests/Persistence/ObjectManagerDecoratorTest.php
         - %currentWorkingDirectory%/tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
         - %currentWorkingDirectory%/tests/Doctrine/Tests/Persistence/Mapping/_files/annotation/TestClass.php
+
     excludes_analyse:
         - %currentWorkingDirectory%/tests/Doctrine/Tests/Persistence/Mapping/_files/TestEntity.php
+
+    ignoreErrors:
+        # looks like a bug in phpstan since method_exists first argument accepts a string
+        - '#Call to function method_exists\(\) with string and (.*) will always evaluate to false#'
```
alcaeus: Where does this error appear? I believe PHPStan pretty much does the `method_exists` call you add to detect any useless calls. This may be an error in type hinting or type documentation.

lib/Doctrine/Persistence/AbstractManagerRegistry.php
```diff
@@ -32,15 +34,17 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
...
+    public function __construct(
+        string $name,
+        array $connections,
+        array $managers,
+        string $defaultConnection,
+        string $defaultManager,
+        string $proxyInterfaceName
+    ) {
         $this->name               = $name;
         $this->connections        = $connections;
```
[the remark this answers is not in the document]
ostrolucky: 👎 On that, people should use static analyzer.

lib/Doctrine/Persistence/ObjectRepository.php
ostrolucky: That's too big BC break. Userspace needs to be able to keep relying on possibility to use array_ functions here

lib/Doctrine/Persistence/AbstractManagerRegistry.php
```diff
@@ -139,77 +141,92 @@ public function getDefaultManagerName()
...
 
         if (! isset($this->managers[$name])) {
-            throw new InvalidArgumentException(sprintf('Doctrine %s Manager named "%s" does not exist.', $this->name, $name));
+            throw new InvalidArgumentException(
+                sprintf('Doctrine %s Manager named "%s" does not exist.', $this->name, $name)
+            );
         }
 
-        return $this->getService($this->managers[$name]);
+        /** @var ObjectManager $objectManager */
```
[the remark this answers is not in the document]
ostrolucky: We should be consistent about this and https://github.com/doctrine/coding-standard/pull/47 didn't pass so far

lib/Doctrine/Persistence/AbstractManagerRegistry.php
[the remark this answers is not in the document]
alcaeus: We can't require people to use a static analyser, which is why I was suggesting to enforce what's being passed into the class.

lib/Doctrine/Persistence/ObjectRepository.php
alcaeus: Ah, usage of array functions was something I hadn't considered. However, I still think it's worth it to allow returning things other than arrays.

lib/Doctrine/Persistence/AbstractManagerRegistry.php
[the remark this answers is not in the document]
ostrolucky: It's not a requirement, you can just respect the contract, no need to be forced into doing that. I'm uneasy on putting such checks which add more paths to our codebase and have performance impact for userspace.

phpstan.neon
jwage: It smells like a bug in phpstan to me. `method_exists` accepts a class name or object instance and PHPStan seems to think it can only accept an object instance. https://github.com/doctrine/persistence/blob/types/lib/Doctrine/Persistence/Mapping/Driver/StaticPHPDriver.php#L134

lib/Doctrine/Persistence/ObjectRepository.php
jwage: I would prefer to make this change in another PR. I want to maintain behavior in this PR as much as possible unless it is out of our control when adding existing documented types.

lib/Doctrine/Persistence/ObjectRepository.php
jwage: We should do this work in another specific PR. Adding new methods instead of changing existing ones would probably be better imo too.

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
jwage: I actually like that rule. The problem imo is caused by grouping lines of code that have no logical reason to be grouped. If you have lines that are grouped together, aligning them makes sense. Example of lines of code that make sense to be grouped and aligned:

```
$something1 = $input->getOption('something1');
$option2    = $input->getOption('option2');
$option3    = $input->getOption('option3');
```

Since they are all the same type of operation, I think it makes sense to group them and align them. I tend to put new lines inbetween statements that don't make sense to logically group so that the code has some breathing room and it is easier to read. Of course, this is subjective and is only how I feel. I don't know if others think this way as well.

lib/Doctrine/Persistence/AbstractManagerRegistry.php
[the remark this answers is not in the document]
jwage: I could go either way. Having more checks and exceptions for end users is good. The performance impact is negligible. I will add it.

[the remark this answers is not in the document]
jwage: Back tracking on this. Not going to add it. PHPStan won't let me do it haha. Not worth spending time on it and blocking this PR from being merged as it is.

phpstan.neon
muglug: Yes, [it's a bug](https://github.com/phpstan/phpstan/issues/1482)

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
jwage: Hmm, I can't remember now. I thought null was being passed before and phpstan was complaining about it, but now I can't find the null. I will undo it.

alcaeus: The null was removed in a bug fix in 1.1.1

jwage: I knew I wasn't crazy :)

jwage: I mean....I am crazy, but not in that specific instance.

lib/Doctrine/Persistence/AbstractManagerRegistry.php
```diff
@@ -32,15 +34,17 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
     private $proxyInterfaceName;
 
     /**
-     * @param string   $name
      * @param string[] $connections
```
Ocramius: Can we improve this to specify the keys, if applicable?

lib/Doctrine/Persistence/AbstractManagerRegistry.php
```diff
@@ -32,15 +34,17 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
...
      * @param string[] $managers
```
Ocramius: Can we improve this to specify the key type, if applicable?

lib/Doctrine/Persistence/AbstractManagerRegistry.php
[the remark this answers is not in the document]
Ocramius: Was gonna comment the same as well: really need someone to pick up doctrine/coding-standard#47 - will shout at the twitters.

lib/Doctrine/Persistence/ConnectionRegistry.php
```diff
@@ -12,28 +14,26 @@ interface ConnectionRegistry
...
-     *
-     * @return object
      */
-    public function getConnection($name = null);
+    public function getConnection(?string $name = null) : object;
 
     /**
      * Gets an array of all registered connections.
      *
      * @return object[] An array of Connection instances.
```
Ocramius: Can we specify the key type here?

lib/Doctrine/Persistence/ConnectionRegistry.php
```diff
@@ -12,28 +14,26 @@ interface ConnectionRegistry
...
      */
-    public function getConnections();
+    public function getConnections() : array;
 
     /**
      * Gets all connection names.
      *
      * @return string[] An array of connection names.
```
Ocramius: Can we specify the key type here?

lib/Doctrine/Persistence/ManagerRegistry.php
```diff
@@ -58,31 +56,30 @@ public function resetManager($name = null);
      *
      * @return string The full namespace.
      */
-    public function getAliasNamespace($alias);
+    public function getAliasNamespace(string $alias) : string;
 
     /**
      * Gets all object manager names.
      *
      * @return string[] An array of object manager names.
```
Ocramius: Probably `array<int, string>` here

lib/Doctrine/Persistence/ConnectionRegistry.php
```diff
@@ -26,14 +26,14 @@ public function getConnection(?string $name = null) : object;
...
-     * @return object[] An array of Connection instances.
+     * @return array<string, object> An array of Connection instances.
      */
     public function getConnections() : array;
 
     /**
      * Gets all connection names.
      *
-     * @return string[] An array of connection names.
+     * @return array<string, string> An array of connection names.
```
Ocramius: Are these really indexed by name?

lib/Doctrine/Persistence/ManagerRegistry.php
```diff
@@ -61,7 +61,7 @@ public function getAliasNamespace(string $alias) : string;
     /**
      * Gets all object manager names.
      *
-     * @return string[] An array of object manager names.
+     * @return array<string, string> An array of object manager names.
```
Ocramius: Similar to the above: is this really indexed by name?

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
```diff
@@ -34,15 +35,17 @@ abstract class FileDriver implements MappingDriver
...
+     * @param string|array<int, string>|FileLocator $locator A FileLocator or one/multiple paths
+     *                                                       where mapping documents can be found.
      */
     public function __construct($locator, ?string $fileExtension = null)
     {
         if ($locator instanceof FileLocator) {
             $this->locator = $locator;
         } else {
-            $this->locator = new DefaultFileLocator((array) $locator, $fileExtension);
+            $paths = is_string($locator) ? [$locator] : $locator;
```
Ocramius: An `(array)` cast was correct here (assuming `array<string>|string`

lib/Doctrine/Persistence/ConnectionRegistry.php
jwage: Yes, I thought it was weird too but I didn't want to change the behavior in this PR.

lib/Doctrine/Persistence/ManagerRegistry.php
jwage: Yes, I thought it was weird too but I didn't want to change the behavior in this PR.

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
jwage: Do you prefer the the casting over this? I thought this might be more desirable/explicit over the cast.

ostrolucky: I prefer array cast. Less branches

Ocramius: A cast is safe in some scenarios, and it is indeed a more efficient operation.

Casting is often risky, but here we know what is going on, IMO

lib/Doctrine/Persistence/Mapping/Driver/StaticPHPDriver.php
```diff
@@ -24,41 +27,41 @@ class StaticPHPDriver implements MappingDriver
...
     private $classNames;
 
     /**
-     * @param string[]|string $paths
+     * @param array<int, string>|string $paths
      */
     public function __construct($paths)
     {
-        $this->addPaths((array) $paths);
+        if (is_string($paths)) {
```
ostrolucky: this is same story

lib/Doctrine/Persistence/Mapping/Driver/DefaultFileLocator.php
```diff
@@ -38,43 +42,45 @@ class DefaultFileLocator implements FileLocator
...
+     * @param string|array<int, string> $paths         One or multiple paths where mapping documents
+     *                                                 can be found.
+     * @param string|null               $fileExtension The file extension of mapping documents,
+     *                                                 usually prefixed with a dot.
      */
-    public function __construct($paths, $fileExtension = null)
+    public function __construct($paths, ?string $fileExtension = null)
     {
-        $this->addPaths((array) $paths);
+        if (is_string($paths)) {
```
ostrolucky: and this

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
jwage: Fixed

lib/Doctrine/Persistence/Mapping/Driver/StaticPHPDriver.php
jwage: Fixed.

lib/Doctrine/Persistence/Mapping/Driver/DefaultFileLocator.php
jwage: Fixed.

PR #51, opened 2019-04-17
Remove PersistentObject magic object.

jwage:
PersistentObject was initially introduced to show that you can do ActiveRecord with Doctrine components.

Its usage has never been endorsed, and objects "saving themselves" are mostly an idea that is better implemented in other tools that focus just on that, while the Doctrine organization mostly focuses on mappers.

If somebody needs this, it will be trivial for them to re-implement this in their own domain, but we shouldn't push it forward from our end.

PR to deprecate this functionality in 1.2 first can be found here https://github.com/doctrine/persistence/pull/55


ostrolucky:
"Why" is needed

Ocramius:
@ostrolucky here's my attempt to clear up the "why":

`PersistentObject` was initially designed as a stunt by @beberlei to show that "yes, you can do ActiveRecord even with Doctrine components".

Its usage has never been endorsed, and objects "saving themselves" are mostly an idea that is better implemented in other tools that focus just on that, while the Doctrine organisation mostly focuses on mappers.

If somebody needs this, it will be trivial for them to re-implement this in their own domain, but we shouldn't push it forward from our end.

ostrolucky:
Thanks, that's all everybody out of the loop needs

jwage:
@ostrolucky PR to add deprecation to 1.2 here https://github.com/doctrine/persistence/pull/55

PR #61, opened 2019-04-24
Add more specific doc block type hints for arrays.

jwage:
While working on https://github.com/doctrine/skeleton-mapper/pull/28/files I found some more types that should be improved.

### DIFF
--- a/lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
+++ b/lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
@@ -124,19 +124,16 @@ abstract class FileDriver implements MappingDriver
             return $this->locator->getAllClassNames($this->globalBasename);
         }
 
-        /** @var ClassMetadata[] $classCache */
+        /** @var array<string, ClassMetadata> $classCache */
         $classCache = $this->classCache;
 
-        /** @var string[] $keys */
+        /** @var array<int, string> $keys */
         $keys = array_keys($classCache);
 
-        /** @var string[] $merged */
-        $merged = array_unique(array_merge(
+        return array_unique(array_merge(
             $keys,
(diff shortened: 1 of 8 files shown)


lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
```diff
@@ -124,19 +124,17 @@ public function getAllClassNames() : array
...
+        /** @var array<string, ClassMetadata> $classCache */
         $classCache = $this->classCache;
 
-        /** @var string[] $keys */
+        /** @var array<int, string> $keys */
         $keys = array_keys($classCache);
 
-        /** @var string[] $merged */
-        $merged = array_unique(array_merge(
+        /** @var array<int, string> $merged */
```
Ocramius: This variable no longer exists.

lib/Doctrine/Persistence/ObjectRepository.php
```diff
@@ -34,10 +34,10 @@ public function findAll() : array;
      * an UnexpectedValueException if certain values of the sorting or limiting details are
      * not supported.
      *
-     * @param mixed[]       $criteria
-     * @param string[]|null $orderBy
+     * @param array<string, mixed> $criteria
+     * @param array<string, mixed> $orderBy
```
Ocramius: This was `string[]`: new format seems wrong

jwage: I changed it to `@param array<string, string> $orderBy`

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
jwage: Removed.

PR #67, opened 2019-08-17
RFC: Add optional orderBy parameter to findAll

maldoinc:
In this patch I want to discuss the possibility of adding an optional parameter to the `findAll` method, that IMO makes sense to be there: `$orderBy`. 

When one needs to fetch the data sorted differently than the default method `findAll` is of no use and one has to use `findBy([], $orderBy)`, which makes it slightly less readable as there are no filters being applied, but rather it is used only for sorting.  Said calls then can be changed to `findAll($orderBy)`

If the maintainers agree that this is a positive change then I can also send the corresponding PR for `doctrine/orm` and any documentation changes.

### DIFF
--- a/lib/Doctrine/Persistence/ObjectRepository.php
+++ b/lib/Doctrine/Persistence/ObjectRepository.php
@@ -27,7 +27,7 @@ interface ObjectRepository
      *
      * @return array<int, object> The objects.
      */
-    public function findAll(?array $orderBy = null) : array;
+    public function findAll(?array $orderBy = []) : array;
 
     /**
      * Finds objects by a set of criteria.
--- a/lib/Doctrine/Persistence/ObjectRepository.php
+++ b/lib/Doctrine/Persistence/ObjectRepository.php
@@ -27,7 +27,7 @@ interface ObjectRepository
      *
      * @return array<int, object> The objects.
      */
-    public function findAll(?array $orderBy) : array;
+    public function findAll(?array $orderBy = null) : array;
 
     /**
      * Finds objects by a set of criteria.
--- a/lib/Doctrine/Persistence/ObjectRepository.php
+++ b/lib/Doctrine/Persistence/ObjectRepository.php
@@ -23,9 +23,11 @@ interface ObjectRepository
     /**
      * Finds all objects in the repository.
      *
+     * @param array<string, string> $orderBy
+     *
      * @return array<int, object> The objects.
      */
-    public function findAll() : array;
+    public function findAll(?array $orderBy) : array;
 
     /**
      * Finds objects by a set of criteria.


alcaeus:
If we make changes to the signature of `findAll`, I would vote for using the same signature we use for the `findBy` method, minus the `$criteria` argument:
```php
public function findAll(array $orderBy = [], $limit = null, $offset = null)
```

Opinions @doctrine/doctrinecore?

maldoinc:
My reasoning for not adding limit and offset is because then it's not technically "all" anymore.

alcaeus:
> My reasoning for not adding limit and offset is because then it's not technically "all" anymore.

🤦‍♂️ Um, yeah. That makes sense.

lib/Doctrine/Persistence/ObjectRepository.php
```diff
@@ -23,9 +23,11 @@ public function find($id) : ?object;
     /**
      * Finds all objects in the repository.
      *
+     * @param array<string, string> $orderBy
+     *
      * @return array<int, object> The objects.
      */
-    public function findAll() : array;
+    public function findAll(?array $orderBy) : array;
```
ostrolucky: It shouldn't be non-optional nullable array, but array with empty array as default. Anyway I'm not sure we will accept this BC break, but I'm pro.

ostrolucky (changes requested):

lib/Doctrine/Persistence/ObjectRepository.php
maldoinc: nullable array seems to be in the signature of `findBy` which is what `findAll` calls in `EntityRepository`.

```php
public function findBy(array $criteria, ?array $orderBy = null, $limit = null, $offset = null)
{
    $persister = $this->em->getUnitOfWork()->getEntityPersister($this->entityName);

    return $persister->loadAll($criteria, $orderBy ?? [], $limit, $offset);
}
```
it also seems that it will default to an empty array regardless.

Anyways, happy to change, just that using the default of `findBy` seemed the sensible thing to do.

Also, unless I'm missing something why would this be a BC? Existing code unaware of this argument will continue to work as before, no?

maldoinc: Regarding the BC stuff, I just realized there wasn't a default value...

ostrolucky: Null default was oversight, please use empty array https://github.com/doctrine/orm/blob/0c2ccec9151fd0b44a60f9cc149e2a107d380799/lib/Doctrine/ORM/EntityRepository.php#L126

But BC break is not just because of that https://3v4l.org/2OnNR

stof: that's a BC break for implementers, not for consumers. But that's still a BC break.

PR #69, opened 2019-11-05
AbstractManagerRegistry: Treat defaultManager first for getManagerForClass

clemdesign:
I propose this PR to fix an issue: when you have two managers, and the default is set as second, the first manager is returned.

### DIFF
--- a/lib/Doctrine/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Persistence/AbstractManagerRegistry.php
@@ -184,6 +184,7 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
         }
 
         if($this->defaultManager) {
+            /** @var ObjectManager $manager */
           $manager = $this->getManager($this->defaultManager);
 
           if (! $manager->getMetadataFactory()->isTransient($class)) {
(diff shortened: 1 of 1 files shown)


malarzm:
This looks odd. Generally speaking same class shouldn't be managed by two (or more) entity managers, that is asking for troubles somewhere down the line. What's the use case?

clemdesign:
I'm use doctrine in Symfony framework and 2 managers because I have implemented a synchronization feature between 2 bases.

The default manager is defined by global parameter, and one times (I'm not call getManager() by defining explicitly the expected manager except for synchronization feature).

During validation process which use an entity, the getManagerForClass() is called and return the first manager in list, whereas the default and expected manager is the second.

malarzm:
I think that for such distinction you should manually pass an entity manager instance to your forms/constraints and rely on it instead of hoping the `getManagerForClass()` will return you the one manager you currently need.

alcaeus:
@malarzm already said all there is to say: the manager registry always assumes that an entity is only managed by a single entity manager. Without that assumption, you'll have to work around issues yourself. Closing as `Won't fix`.

PR #71, opened 2019-12-06
Namespace move BC compatibility layer

greg0ire:
This is just a draft, tell me what you think about the example, and I will squash my commits and apply it to all other classes.
Please tell me what should be done about the tests. I assumed they should use the new names, but tell me if you think previous tests with the old names should be kept in a `Legacy` subdirectory.

The corresponding commit on master is fed87ebed584609c5e62749c99cfb27ac911c169


lib/Doctrine/Common/NotifyPropertyChanged.php
```diff
@@ -2,19 +2,24 @@
...
- */
-interface NotifyPropertyChanged
-{
+if (!interface_exists(\Doctrine\Persistence\NotifyPropertyChanged::class, false)) {
+    @trigger_error(
+        'The '.__NAMESPACE__.'\NotifyPropertyChanged class is deprecated since doctrine/persistence 1.2 and will be removed in 2.0.'
+        .' Use \Doctrine\Persistence\NotifyPropertyChanged instead.',
+        E_USER_DEPRECATED
+    );
+}
```
greg0ire: Are you ok with this or do you think line 20 should be enough?

alcaeus:
Heads up, I’ll push a 1.3.x branch for this to target. We’ll add more compatibility layers in there for the BC breaks in 2.0.

lib/Doctrine/Common/NotifyPropertyChanged.php
```diff
@@ -2,19 +2,24 @@
...
- *
- * This interface is implemented by objects that manually want to notify their object manager or
- * other listeners when properties change, instead of relying on the object manager to compute
- * property changes itself when changes are to be persisted.
- */
-interface NotifyPropertyChanged
-{
+if (!interface_exists(\Doctrine\Persistence\NotifyPropertyChanged::class, false)) {
+    @trigger_error(
+        'The '.__NAMESPACE__.'\NotifyPropertyChanged class is deprecated since doctrine/persistence 1.2 and will be removed in 2.0.'
```
alcaeus:
```suggestion
        'The "'.__NAMESPACE__.'\NotifyPropertyChanged" interface is deprecated since doctrine/persistence 1.3 and will be removed in 2.0. Please use "..." instead.'
```

We could also use `sprintf` instead of string concatenation here.

lib/Doctrine/Common/NotifyPropertyChanged.php
```diff
@@ -2,19 +2,24 @@
...
+    \Doctrine\Persistence\NotifyPropertyChanged::class,
+    __NAMESPACE__.'\NotifyPropertyChanged'
+);
+
+if (false) {
     /**
-     * Adds a listener that wants to be notified about property changes.
-     *
-     * @return void
+     * @deprecated 1.2 Use Doctrine\Persistence\NotifyPropertyChanged
```
alcaeus:
```suggestion
     * @deprecated 1.3 Use Doctrine\Persistence\NotifyPropertyChanged
```

lib/Doctrine/Common/NotifyPropertyChanged.php
alcaeus: If the DebugClassLoader already triggers a deprecation based on the `@deprecated` annotation we should skip this to avoid duplicate deprecation messages. Otherwise, I definitely want to keep these in.

// cc @nicolas-grekas

nicolas-grekas: That practice on Symfony is to still add all possible runtime deprecations, because using the DebugClassLoader is optional, especially when using packages standalone.

alcaeus: That clears it up. Thanks Nicolas!

@greg0ire I would suggest we add this deprecation message to all deprecated classes/interfaces then.

greg0ire:
Travis fails because of `Abstract` / `Exception` on classes that are considered new for phpcs but actually aren't

greg0ire:
I checked that tests that used all class names still work, you can do too by checkout out this PR, then running `git checkout origin/1.3.x tests && vendor/bin/phpunit`

greg0ire:
I fixed conflict and reverted the move of `PersistentObject`… that class is not present on master, thanks @alcaeus for thinking of that!

alcaeus:
@greg0ire feel free to add exceptions to phpcs.xml for the current naming errors. These should be taken care of separately - I don't feel comfortable shoving that into 2.0 at this point.

greg0ire:
Oh right, I assumed the phpcs was enabled on touched/new files only somehow. I pushed a separate commit.

alcaeus:
Thanks @greg0ire!

nicolas-grekas:
Tests on Symfony fail with deprecations after this change so I'm on a PR to fix them.
But it happens that's not possible: the deprecation layer in the PR doesn't work I think.

The issue is that class_alias doesn't work, because if a type hint uses the legacy one, no deprecation will be triggered - and the alias won't ever be loaded - but PHP will fail with a fatal error saying "incompatible type hint".

Fixing this would require loading the alias all the time (and this silencing the deprecation, which is fine if there is no better way...)

alcaeus:
Thanks for the feedback! I’ll try to work on this today.

nicolas-grekas:
See https://github.com/symfony/symfony/pull/34949

nicolas-grekas:
Actually the aliases are always loaded, my bad. I made 3.4 green. Let's try with upper branches.

Majkl578:
Honestly this approach is disgusting as it totally breaks preloading.

Majkl578:
cc @Ocramius @lcobucci @beberlei @goetas This is serious with PHP 7.4 being out.

alcaeus:
This is perfectly fine if our aim is to not break the entire ecosystem. If preloading breaks, you’re welcome to create a PR to fix it without removing the deprecation layer. Thanks.

PR #75, opened 2019-12-11
Remove return types from all interfaces

alcaeus:
This partially reverts commit 3e2fad57e1f180e9d6f827c17548171959c1a9a7.

To avoid a hard BC break from 1.x to 2.0, we remove the milestone, as that allows people to install either 1.x or 2.0.

With this, the upgrade process is as follows:
* 1.3.x will deprecate the `Doctrine\Common\Persistence` namespace in favour of the `Doctrine\Persistence` namespace.
* Once the above has been changed, a project can change its version constraint for doctrine/persistence to `^1.3 || ^2.0` without further changes.
* Projects now have to add return type hints to all methods extended from persistence interfaces and classes.
* After a while, we will release version 3.0 which will require return type hints, but will otherwise be compatible with 2.x. This will allow projects to change the version constraint to `^2.0 || ^3.0`.

This ensures a smooth transition where the ecosystem has time to catch up. Otherwise, a single package that upgrades to doctrine/persistence 2.0 could fast-travel a user into dependency hell. **Projects are discouraged from running any single major version constraint** - e.g. running `^2.0` or `^1.0` is discouraged. Dependents should quickly ensure compatibility with 2.0 and update the version constraint as suggested above.

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -17,4 +17,5 @@ ActiveRecord style functionality.
 
 ## Strict Typing
 
-Proper type hints have been added to all method signatures and strict types have been enabled.
+Type declarations have been enabled for all parameters and strict types have been enabled.
+Return type declarations will be added in the next major release.
--- a/lib/Doctrine/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Persistence/AbstractManagerRegistry.php
@@ -60,9 +60,9 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
      *
      * @param string $name The name of the service.
      *
-     * @return object The instance of the given service.
+     * @return ObjectManager The instance of the given service.
      */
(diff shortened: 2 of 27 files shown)


UPGRADE-2.0.md
```diff
@@ -17,4 +17,5 @@ ActiveRecord style functionality.
 
 ## Strict Typing
 
-Proper type hints have been added to all method signatures and strict types have been enabled.
+Type hints have been enabled for all parameters and strict types have been enabled.
```
greg0ire:
```suggestion
Type declarations have been enabled for all parameters and strict types have been enabled.
```

UPGRADE-2.0.md
```diff
@@ -17,4 +17,5 @@ ActiveRecord style functionality.
...
+Return type hints will be added in the next major release.
```
greg0ire:
```suggestion
Return type declarations will be added in the next major release.
```

PR #79, opened 2019-12-12
Add missing autoload calls

### LINKED ISSUE #78: Compile Error: Declaration of Doctrine XmlDriver
Symfony 4.4, on composer u i got this:

  - Updating doctrine/persistence (1.2.0 => 1.3.0): Loading from cache
!!  In XmlDriver.php line 881:
!!                                                                                 
!!    Compile Error: Declaration of Doctrine\ORM\Mapping\Driver\XmlDriver::loadMe  
!!    tadataForClass($className, Doctrine\Common\Persistence\Mapping\ClassMetadat  
!!    a $metadata) must be compatible with Doctrine\Persistence\Mapping\Driver\Ma  
!!    ppingDriver::loadMetadataForClass($className, Doctrine\Persistence\Mapping\  
!!    ClassMetadata $metadata)

greg0ire:
Types extending these but having the old type declarations might not be
considered compatible unless we autoload the type declarations.

Fixes #78 

# How can I test this?

```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/doctrine-persistence
composer require doctrine/persistence "dev-load-classes-in-type-declarations as 1.3.0"
```

### DIFF
--- a/lib/Doctrine/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Persistence/AbstractManagerRegistry.php
@@ -6,6 +6,7 @@ use InvalidArgumentException;
 use ReflectionClass;
 use function class_exists;
 use function explode;
+use function interface_exists;
 use function sprintf;
 use function strpos;
 
@@ -247,3 +248,4 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
 }
 
 class_exists(\Doctrine\Common\Persistence\AbstractManagerRegistry::class);
+interface_exists(\Doctrine\Common\Persistence\ObjectManager::class);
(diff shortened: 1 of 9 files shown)


lib/Doctrine/Persistence/Event/LifecycleEventArgs.php
```diff
@@ -61,3 +62,4 @@ public function getObjectManager()
 }
 
 class_exists(\Doctrine\Common\Persistence\Event\LifecycleEventArgs::class);
+interface_exists(\Doctrine\Common\Persistence\ObjectManager::class);
```
greg0ire: That one may be useless because it refers to the constructor, no signature compatibility issue should happen here.

Issue #81, opened 2019-12-13
doctrine/persistence 1.3.2 bug


itarcontact:
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Doctrine\ORM\Configuration::setMetadataDriverImpl() must be an instance of Doctrine\Common\Persistence\Mapping\Driver\MappingDriver, instance of Doctrine\Persistence\Mapping\Driver\MappingDriverChain given

mikaelcom:
> PHP Fatal error: Uncaught TypeError: Argument 1 passed to Doctrine\ORM\Configuration::setMetadataDriverImpl() must be an instance of Doctrine\Common\Persistence\Mapping\Driver\MappingDriver, instance of Doctrine\Persistence\Mapping\Driver\MappingDriverChain given

I just faced the issue too :cry:

TomasVotruba:
Same here

itarcontact:
I updated the symfony project today with a composer. Updated version of doctrine/persistence 1.3.1 with many deprecations. And now this bug with 1.3.2 version.

TeQno:
same here

michaljusiega:
https://github.com/doctrine/persistence/pull/80

The guys from doctrine are funny to me ...

aguvillalba:
Same issue here in combination with laravel-doctrine package:

Argument 1 passed to Doctrine\ORM\Configuration::setMetadataDriverImpl() must be an instance of Doctrine\Common\Persistence\Mapping\Driver\MappingDriver, instance of Doctrine\ORM\Mapping\Driver\XmlDriver given, called in /code/vendor/laravel-doctrine/orm/src/EntityManagerFactory.php on line 167

alexander-schranz:
You can add the broken version as a conflict to your `composer.json`:

```
{
    "conflict": {
        "doctrine/persistence": "1.3.2"
    }
}
```

Then 1.3.3 will install in future which I think will include definitly a fix for this issue.

alcaeus:
1.3.3 has been released. I apologise for the problems. I’ll post a blog post about what we’re doing and why it failed later today or after the weekend. Thanks for understanding.

Toflar:
[...]

PR #82, opened 2019-12-13
Fix BC layer

nicolas-grekas:
As spotted in https://github.com/symfony/symfony/pull/34955#issuecomment-565381150


ro0NL:
see https://github.com/doctrine/persistence/issues/81

PR #87, opened 2020-01-08
Temporarily remove deprecations about ns move

greg0ire:
We want to give ourselves time to address them in the other packages we
own and that consume this library.
Not doing so generates a lot of support from people asking how they can
fix the deprecations in their application, which they cannot. This
commit should be reverted in a following release, when at least all of
our packages have made the move.


greg0ire:
To be merged after discussing this in https://github.com/orgs/doctrine/teams/doctrinecore/discussions/11

Issue #89, opened 2020-01-09
Add support TypedNoDefaultReflectionProperty class to RuntimeReflectionService


sspat:
As suggested by @beberlei in doctrine/orm#7857
This is part of the implementation to fix this bug.

sspat:
@greg0ire Could you please advise me on what version branch I should target with the implementation?

greg0ire:
The lowest maintained branch is 1.2.x please target the lowest _affected_ branch (no idea which it is)

beberlei:
Fixed

PR #90, opened 2020-01-09
Added support for TypedNoDefaultReflectionProperty class to RuntimeReflectionService

sspat:
Implementation for issue #89 
See discussion in doctrine/orm#7857

As suggested by @beberlei, the `Doctrine\Common\Reflection\TypedNoDefaultReflectionProperty` class was introduced and the `Doctrine\Persistence\Mapping\RuntimeReflectionService::getAccessibleProperty` returns an instance of this class if the property is typed and has no default value.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/reflection": "^1.0"
+        "doctrine/reflection": "^1.1"
     },
     "require-dev": {
         "phpstan/phpstan": "^0.11",
--- a/lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
+++ b/lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
@@ -5,10 +5,12 @@ declare(strict_types=1);
 namespace Doctrine\Persistence\Mapping;
 
 use Doctrine\Common\Reflection\RuntimePublicReflectionProperty;
+use Doctrine\Common\Reflection\TypedNoDefaultReflectionProperty;
 use ReflectionClass;
 use ReflectionException;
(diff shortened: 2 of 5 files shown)


sspat:
Also updated phpunit config to allow running tests with php 7.4 syntax.

.travis.yml
```diff
@@ -49,6 +49,7 @@ jobs:
       script: ./vendor/bin/phpcs
 
     - stage: Code Quality
+      php: 7.4snapshot
```
greg0ire: A stable image is out
```suggestion
      php: 7.4
```

composer.json
```diff
@@ -25,7 +25,7 @@
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/reflection": "^1.0"
+        "doctrine/reflection": "^1.1"
```
alcaeus: With the dependency change, this will only be in 1.4 which will not be released until we've fixed the deprecation warnings in ORM and other libraries. See my other comment for a suggestion on how to fix this.

lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
```diff
@@ -64,7 +66,9 @@ public function getAccessibleProperty(string $class, string $property)
     {
         $reflectionProperty = new ReflectionProperty($class, $property);
 
-        if ($reflectionProperty->isPublic()) {
+        if (! array_key_exists($property, $this->getClass($class)->getDefaultProperties())) {
```
alcaeus: With this change, you can avoid the dependency bump and target 1.3 to fix this bug in a patch release:

```suggestion
        if (class_exists(TypedNoDefaultReflectionProperty::class) && ! array_key_exists($property, $this->getClass($class)->getDefaultProperties())) {
```

PR #94, opened 2020-01-16
Add typed no default reflection property class usage

### LINKED ISSUE #89: Add support TypedNoDefaultReflectionProperty class to RuntimeReflectionService

beberlei:
Supersedes #90 

This adds a check for PHP 7.4 so that we don't incur the overhead of `getDefaultProperties` every time.

It also changes the order of the reflection property creations, because RuntimePublicReflectionProperty already works with typed no default properties, and its the more specialized version that is necessary to be used to don't break lazy loading on typed properties.

@sspat Can you check?

Fixes #89

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
-  - 7.4snapshot
+  - 7.4
 
 cache:
   directories:
@@ -23,9 +23,6 @@ script:
   - ./vendor/bin/phpunit
 
 jobs:
-  allow_failures:
-    - php: 7.4snapshot
-
   include:
     - stage: Test
       env: DEPENDENCIES=low
@@ -50,6 +47,7 @@ jobs:
       script: ./vendor/bin/phpcs
 
     - stage: Code Quality
+      php: 7.4
       env: STATIC_ANALYSIS
       install: travis_retry composer install --prefer-dist
       script: vendor/bin/phpstan analyze
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/reflection": "^1.0"
+        "doctrine/reflection": "^1.1"
     },
     "require-dev": {
         "phpstan/phpstan": "^0.11",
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/reflection": "^1.0"
+        "doctrine/reflection": "^1.1"
```
alcaeus: Noting that this is required for a fix. Since this transitive dependency fixes the issue for ORM, we've decided to bump this in a patch release.

lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
```diff
@@ -5,18 +5,31 @@
...
  * PHP Runtime Reflection Service.
  */
 class RuntimeReflectionService implements ReflectionService
 {
+    /** @var bool */
+    private $supportsTypedPropertiesWorkaround;
+
+    public function __construct()
+    {
+        $this->supportsTypedPropertiesWorkaround =
```
alcaeus: I don't recall us having a rule on line length, does phpcs complain if you put the code in one line? If not, I wouldn't wrap this.

beberlei:
@sspat its going into 1.3

.travis.yml
```diff
@@ -24,7 +24,7 @@ script:
 
 jobs:
   allow_failures:
-    - php: 7.4snapshot
+    - php: 7.4
```
alcaeus: Making a note that we should remove this section before releasing the next patch release since we're explicitly targeting PHP 7.4 with it.

beberlei: I just removed it now, otherwise we forget anyways and since its not failing anyways its just a minor cleanup.

alcaeus: Ah, I was expecting it to fail. If it's passing, we're good 🎉

greg0ire:
I got rid of the merge commit and squashed the "address Andreas review" commit into the first

lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
```diff
@@ -3,18 +3,30 @@
...
  * PHP Runtime Reflection Service.
  */
 class RuntimeReflectionService implements ReflectionService
 {
+    /** @var bool */
+    private $supportsTypedPropertiesWorkaround;
+
+    public function __construct()
+    {
+        $this->supportsTypedPropertiesWorkaround = version_compare((string) phpversion(), '7.4.0') >= 0;
```
stof: I suggest replacing this property with an inline `\PHP_VERSION_ID > 70400` check, which will be resolved at compile time by OPCache.

PR #100, opened 2020-03-21
Bump doctrine/reflection to 1.2

beberlei:
Includes a bugfix for PHP 7.4 and typed properties, no backwards compatibility problems.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/reflection": "^1.1"
+        "doctrine/reflection": "^1.2"
     },
     "require-dev": {
         "phpstan/phpstan": "^0.11",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/reflection": "^1.1"
+        "doctrine/reflection": "^1.2"
```
greg0ire: `^1.1` already allows installing 1.2.0 and higher. Is your goal to make sure 1.1 can no longer be installed in new versions of persistence?

beberlei: @greg0ire no my goal is that users automatically get the latest version. We had a few people report the same bugs again in ORM even though the fix is in doctrine/reflection, which they didn't upgrade to the latest version.

greg0ire: Ok, not sure this will help a lot, but it can't hurt 👍

PR #103, opened 2020-04-03
Fix incorrect instance of ReflectionProperty created for public typed properties

### LINKED ISSUE #102: Incorrect instance of ReflectionProperty created for public typed properties
https://github.com/doctrine/persistence/blob/f5cf0d59f5c8e451186f70145da224c269be20c8/lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php#L77

Related to: https://github.com/doctrine/orm/issues/7999

The issue was fixed for private properties, but public properties are still falling in the first condition, and being set as null when deleting an entity.

joaojacome:
Resolves #102

### DIFF
--- a/lib/Doctrine/Persistence/Reflection/TypedNoDefaultReflectionPropertyBase.php
+++ b/lib/Doctrine/Persistence/Reflection/TypedNoDefaultReflectionPropertyBase.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Doctrine\Persistence\Reflection;
+
+use Closure;
+use ReturnTypeWillChange;
+
+use function assert;
+
+/**
+ * PHP Typed No Default Reflection Property Base - special override for typed properties without a default value.
+ */
+trait TypedNoDefaultReflectionPropertyBase
+{
+    /**
+     * {@inheritDoc}
+     *
+     * Checks that a typed property is initialized before accessing its value.
+     * This is necessary to avoid PHP error "Error: Typed property must not be accessed before initialization".
+     * Should be used only for reflecting typed properties without a default value.
+     *
+     * @param object $object
+     */
+    #[ReturnTypeWillChange]
+    public function getValue($object = null)
+    {
+        return $object !== null && $this->isInitialized($object) ? parent::getValue($object) : null;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Works around the problem with setting typed no default properties to
+     * NULL which is not supported, instead unset() to uninitialize.
+     *
+     * @link https://github.com/doctrine/orm/issues/7999
+     *
+     * @param object $object
+     */
+    #[ReturnTypeWillChange]
+    public function setValue($object, $value = null)
+    {
+        if ($value === null && $this->hasType() && ! $this->getType()->allowsNull()) {
+            $propertyName = $this->getName();
+
+            $unsetter = function () use ($propertyName): void {
+                unset($this->$propertyName);
+            };
+            $unsetter = $unsetter->bindTo($object, $this->getDeclaringClass()->getName());
+
+            assert($unsetter instanceof Closure);
+
+            $unsetter();
+
+            return;
+        }
+
+        parent::setValue($object, $value);
+    }
+}
--- a/lib/Doctrine/Persistence/Reflection/TypedNoDefaultRuntimePublicReflectionProperty.php
+++ b/lib/Doctrine/Persistence/Reflection/TypedNoDefaultRuntimePublicReflectionProperty.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Doctrine\Persistence\Reflection;
+
+/**
+ * PHP Typed No Default Runtime Public Reflection Property - special override for public typed properties without a default value.
+ */
+class TypedNoDefaultRuntimePublicReflectionProperty extends RuntimePublicReflectionProperty
+{
+    use TypedNoDefaultReflectionPropertyBase;
+}
--- a/tests/Doctrine/Tests_PHP74/Persistence/Mapping/RuntimeReflectionServiceTest.php
+++ b/tests/Doctrine/Tests_PHP74/Persistence/Mapping/RuntimeReflectionServiceTest.php
@@ -7,6 +7,7 @@ namespace Doctrine\Tests_PHP74\Persistence\Mapping;
 use Doctrine\Persistence\Mapping\RuntimeReflectionService;
 use Doctrine\Persistence\Reflection\RuntimePublicReflectionProperty;
 use Doctrine\Persistence\Reflection\TypedNoDefaultReflectionProperty;
+use Doctrine\Persistence\Reflection\TypedNoDefaultRuntimePublicReflectionProperty;
 use PHPUnit\Framework\TestCase;
 use ReflectionProperty;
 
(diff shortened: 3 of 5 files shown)


SenseException:
Thanks for your contribution. Your switch in the if-block is supposed to fix an issue with the created ReflectionProperty instance. Can you please create a test that covers this issue and prevents regression in this project?

joaojacome:
I've added the tests for this issue. Can you check again, please?

tests/Doctrine/Tests_PHP74/Persistence/Mapping/RuntimeReflectionServiceTest.php
```diff
@@ -44,10 +54,43 @@ public function testGetTypedDefaultReflectionProperty() : void
...
+    }
+
+    public function testGetNonTypedDefaultReflectionProperty() : void
+    {
+        $reflProp = $this->reflectionService->getAccessibleProperty(self::class, 'nonTypedDefaultProperty');
+        self::assertInstanceOf(ReflectionProperty::class, $reflProp);
+        self::assertNotInstanceOf(TypedNoDefaultReflectionProperty::class, $reflProp);
+    }
+
     public function testGetTypedPublicNoDefaultPropertyWorksWithGetValue() : void
```
SenseException, line 70: This test got changed by your PR. If the old test using `RuntimePublicReflectionProperty` would fail with your changes, you might have implemented a BC-break here and 1.3.x wouldn't be the appropriate target branch.

oalphenaar:
Will this be merged? As of today it is still impossible to remove entities where the id property is public and this seems to be the fix.

malarzm:
There's https://github.com/doctrine/persistence/pull/103#discussion_r419695194 that needs to be checked/addressed

tests/Doctrine/Tests_PHP74/Persistence/Mapping/RuntimeReflectionServiceTest.php
joaojacome: The old test fails with the fix - but wouldn't the old behaviour be considered a bug?

SenseException: A bugfix that introduces BC break could only be merged into a major release. The fix should also be reviewed by other maintainers too.

greg0ire:
Please associate your email address with your Github account, or change the
email in your commits to an address already associated with it. If you do not
want to expose your personal email address, you may use
`$userName@users.noreply.github.com`, that way we can still reach you through Github.
Additionally, you will get credit for this contribution on your Github profile.

tests/Doctrine/Tests_PHP74/Persistence/Mapping/RuntimeReflectionServiceTest.php
guilliamxavier: This fix has been hold off for too long... as it is we might not even have it for version 3.0... Wouldn't it be possible to preserve BC? e.g. use a new class that extends RuntimePublicReflectionProperty but uses the same code as TypedNoDefaultReflectionProperty?

joaojacome: Thanks for the suggestion. To avoid code duplication, I've added a trait to hold the common methods, although I'm not sure if it's the best approach.

The test now covers the old and the new behaviours.

guilliamxavier: Well my question was primarily to @SenseException (and other maintainers), but I was thinking to something like that yes 🙂 Thanks, now let's wait for new reviews 😉

SenseException (dismissed):

greg0ire:
My review has been requested. Before I review this, I obviously have to read #102, which mentions an incorrect type but does not explain what makes it incorrect, and refers me to doctrine/orm#7999.
doctrine/orm#7999 is a lengthy issue with 17 replies, a wall of text and 11 participants. It mentions doctrine/orm#7940 a pull request with 35 replies, which fixes doctrine/orm#7864, which is another wall of text.

I might find the time and energy to review this one day, but today is not that day, sorry.

If you want to speed things up, please squash your commits into one, and improve your commit message according to [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches). It would be great if it mentioned the reason you are doing this, https://github.com/doctrine/persistence/pull/103#issuecomment-752984270 , what changes you made and why they fix the issue.

guilliamxavier:
@greg0ire: I can at least recap what led me here: The original issue is indeed doctrine/orm#7999: you get a `TypeError` (from deep in Doctrine's internals) when you try to delete an entity with a *generated* id on a *non-nullable* typed property. E.g.:

```php
/** @ORM\Entity */
class MyEntity {
    /** @ORM\Id @ORM\GeneratedValue @ORM\Column(type="integer") */
    public int $id;
}

$myEntity = $entityManager->find(MyEntity::class, 42); // let's assume it exists

$entityManager->remove($myEntity);
$entityManager->flush(); // Typed property Entity::$id must be int, null used
```

(because Doctrine tries to "reset" the id to `null`).

Then doctrine/reflection#35 fixed it but *only for `private`/`protected` properties, not `public`* (maybe because #94 thought that "RuntimePublicReflectionProperty already works with typed no default properties", only considering lazy loading at that time).

Now this PR aims to fix it also for public properties (note: the first fix was simply to reverse the order, resulting in what was initially proposed in #90, but since eventually #94 chose the "wrong" order, there was some BC-change requested).

(As for the EXTRA_LAZY issue, it actually seems unrelated.)

But indeed, there is place for improvement of the commit message (@joaojacome 😉)

greg0ire:
Thanks, I will try to look at this today, it looks a bit more approachable now :)

tests/Doctrine/Tests_PHP74/Persistence/Mapping/RuntimeReflectionServiceTest.php
```diff
@@ -45,6 +55,33 @@ public function testGetTypedPublicNoDefaultPropertyWorksWithGetValue(): void
     {
         $reflProp = $this->reflectionService->getAccessibleProperty(self::class, 'typedNoDefaultPublicProperty');
         self::assertInstanceOf(RuntimePublicReflectionProperty::class, $reflProp);
+        self::assertInstanceOf(TypedNoDefaultRuntimePublicReflectionProperty::class, $reflProp);
```
alcaeus, line 58: Original assertion can be dropped
```suggestion
        self::assertInstanceOf(TypedNoDefaultRuntimePublicReflectionProperty::class, $reflProp);
```

alcaeus (dismissed):
Nit in tests, but LGTM.

tests/Doctrine/Tests_PHP74/Persistence/Mapping/RuntimeReflectionServiceTest.php
guilliamxavier: It was kept on purpose, to ensure that there is no "BC break"

lib/Doctrine/Persistence/Reflection/TypedNoDefaultReflectionPropertyBase.php
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Doctrine\Persistence\Reflection;
+
+use Closure;
+use ReturnTypeWillChange;
+
+use function assert;
+
+/**
+ * PHP Runtime Reflection Public Property - special override for typed properties without a default value.
```
guilliamxavier:
```suggestion
 * PHP Typed No Default Reflection Property Base - special override for typed properties without a default value.
```

lib/Doctrine/Persistence/Reflection/TypedNoDefaultRuntimePublicReflectionProperty.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Doctrine\Persistence\Reflection;
+
+/**
+ * PHP Runtime Reflection Public Property - special override for public typed properties without a default value.
```
guilliamxavier:
```suggestion
 * PHP Typed No Default Runtime Public Reflection Property - special override for public typed properties without a default value.
```

tests/Doctrine/Tests_PHP74/Persistence/Mapping/RuntimeReflectionServiceTest.php
alcaeus: I see. Feel free to leave it in then 👍

joaojacome:
> @greg0ire: 
> But indeed, there is place for improvement of the commit message (@joaojacome wink)

I'll squash it and give it a proper message :)

greg0ire (approved):
Now _that's_ a commit message! It feels a lot safer to merge this, thanks a lot!

greg0ire:
Thanks everyone!

PR #104, opened 2020-05-01
Adds generics annotations

### LINKED ISSUE #72: Add support for generics via static analysis
As was done in doctrine/collections, we should add support for generics using the `@psalm-template` annotations for generic classes like Repositories and Registries. We should also ensure that generics in PHPStan are supported as well.

See https://github.com/doctrine/collections/pull/177 for the reference implementation in collections.

weph:
Not sure if that's enough to close #72 but it's a start ;)

### DIFF
--- a/lib/Doctrine/Persistence/ObjectManager.php
+++ b/lib/Doctrine/Persistence/ObjectManager.php
@@ -20,6 +20,10 @@ interface ObjectManager
      * @param mixed  $id        The identity of the object to find.
      *
      * @return object|null The found object.
+     *
+     * @template T
+     * @psalm-param class-string<T> $className
+     * @psalm-return T|null
      */
     public function find($className, $id);
 
@@ -111,9 +115,9 @@ interface ObjectManager
     /**
      * Gets the repository for a class.
      *
-     * @param string $className
-     *
-     * @return ObjectRepository
+     * @template T
+     * @psalm-param class-string<T> $className
+     * @psalm-return ObjectRepository<T>
      */
     public function getRepository($className);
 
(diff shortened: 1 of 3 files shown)


Ocramius:
`findAll` and `findBy` should return `list<T>`, I think
Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

weph:
Yeah, wasn't sure about that. It's a list in doctrine/orm, but doctrine/mongodb-odm uses the document id as array key. Not sure about the other odm implementations. I agree that it should be a list, though. Overriding the annotations in odm doesn't seem right but that's just my humble opinion.

greg0ire:
Nice PR! You gave me the motivation to setup Psalm on this lib! See #105

weph:
Do you have any opinions on this one, @alcaeus ?

alcaeus:
> Do you have any opinions on this one, @alcaeus ?

Sorry for the delay. Since ODM doesn't return a list but rather a struct, this has to be `T[]` for the time being.

alcaeus:
Note that this should go into 1.4.x since that version is supposed to have feature parity with 2.0.

@weph could you please change the target branch and rebase? Thanks!

greg0ire:
Psalm support just landed on 1.4.x BTW :)

greg0ire:
I did the rebase for you.

greg0ire:
Reopening to get a report from Travis

greg0ire:
Thanks @weph !

weph:
Thank you, @greg0ire !

lib/Doctrine/Persistence/ObjectManager.php
```diff
@@ -111,9 +115,9 @@ public function flush();
     /**
      * Gets the repository for a class.
      *
-     * @param string $className
```
andrew-demb, line 114: I think this line should be preserved to allow IDE know about required argument type.

alcaeus: I agree. Do you mind creating a pull request?

andrew-demb: Here it is #111

PR #105, opened 2020-05-01
Static Analysis with Psalm

greg0ire:


greg0ire:
The action will be visible only after this is merged, but you can see how it looks at https://github.com/greg0ire/persistence/pull/1

greg0ire:
@localheinz @bendavies please review

greg0ire:
I force pushed to add `cs2pr`

Ocramius:
LGTM
Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          path: "~/.composer/cache"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Psalm"
+        run: "vendor/bin/psalm --output-format=checkstyle | cs2pr"
```
bendavies, line 39: psalm has native support, no need for `cs2pr`
`vendor/bin/psalm --show-info=false --stats --output-format=github --threads=4`

psalm.xml
```diff
@@ -0,0 +1,16 @@
...
+<psalm
+    totallyTyped="false"
+    errorLevel="5"
+    resolveFromConfigFile="true"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="https://getpsalm.org/schema/config"
+    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+>
+    <projectFiles>
+        <directory name="lib/Doctrine/Persistence" />
```
bendavies, line 11: not tests?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,39 @@
+
+name: "Continuous Integration"
+
+on:
+  pull_request:
```
bendavies, line 5: no CI runs on branches?

greg0ire: What would be the point of that?

psalm.xml
greg0ire: Woops, didn't mean to leave those out

.github/workflows/continuous-integration.yml
bendavies: do you require PRs to be up to date before merging?
if not, then a passing PR could fail after merging to master?

greg0ire: We don't, and I hadn't thought of that possibility. That being said, it will not prevent the PR from being merged.

psalm.xml
greg0ire: Actually, this will be hard to do until 2.0 and until we use phpunit 8, we should probably leave those out.

greg0ire: I'm giving a try in case that's not that big of an issue.

greg0ire: @bendavies I managed to make it work!  Locally. And now the build Travis fails because phpunit 8 requires php 7.2. So I'm removing these commits, but they are still on a branch on my machine :see_no_evil:

.github/workflows/continuous-integration.yml
localheinz, line 5:
```suggestion
  pull_request: null
```

.github/workflows/continuous-integration.yml
greg0ire: Fixed

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Psalm"
```
localheinz, line 38:
```suggestion
      - name: "Run a static analysis with vimeo/psalm"
```

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,39 @@
...
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+          tools: cs2pr
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
```
bendavies, line 33: add `php-` as discussed?

psalm.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<psalm
+    totallyTyped="false"
+    errorLevel="5"
```
bendavies, line 4: any more as per other projects?

greg0ire: Lower is better, and we can have lvl 4 with 0 modifications, so let's do that.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,38 @@
+
+name: "Continuous Integration"
+
+on:
+  pull_request: null
+
+jobs:
+  static-analysis-psalm:
+    name: "Static Analysis with Psalm"
+    runs-on: ubuntu-18.04
```
carusogabriel, line 10:
```suggestion
    runs-on: ubuntu-latest
```

As previously discussed in other PRs :)

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
```
carusogabriel, line 35: Let's remove `--no-suggest`, Composer v2 doesn't support it.
```suggestion
        run: "composer install --no-interaction --no-progress"
```

.github/workflows/continuous-integration.yml
SenseException: One PR is using latest and another one 18.04 for static analysis? Or is here a different reason than in doctrine/orm#8116 (or it was just forgotten :-D).

greg0ire: We settled for `latest`, just like in ORM, I am going to update this soon

PR #122, opened 2020-07-29
Import useful reflection classes

greg0ire:
Other reflection classes do not seem to be used anywhere.

Classes were imported from the master branch, I hope that is ok?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,12 +24,12 @@
         "doctrine/annotations": "^1.0",
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
-        "doctrine/event-manager": "^1.0",
-        "doctrine/reflection": "^1.2"
+        "doctrine/event-manager": "^1.0"
     },
     "require-dev": {
         "phpstan/phpstan": "^0.11",
         "doctrine/coding-standard": "^6.0",
+        "doctrine/common": "^3.0",
         "phpunit/phpunit": "^7.0 || ^8.0 || ^9.0",
         "vimeo/psalm": "^3.11"
     },
--- a/lib/Doctrine/Persistence/Reflection/RuntimePublicReflectionProperty.php
+++ b/lib/Doctrine/Persistence/Reflection/RuntimePublicReflectionProperty.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Doctrine\Persistence\Reflection;
+
+use Doctrine\Common\Proxy\Proxy;
+use ReflectionProperty;
+
+/**
+ * PHP Runtime Reflection Public Property - special overrides for public properties.
+ */
+class RuntimePublicReflectionProperty extends ReflectionProperty
+{
+    /**
+     * {@inheritDoc}
+     *
+     * Checks is the value actually exist before fetching it.
+     * This is to avoid calling `__get` on the provided $object if it
+     * is a {@see \Doctrine\Common\Proxy\Proxy}.
+     */
+    public function getValue($object = null)
+    {
+        $name = $this->getName();
+
+        if ($object instanceof Proxy && ! $object->__isInitialized()) {
+            $originalInitializer = $object->__getInitializer();
+            $object->__setInitializer(null);
+            $val = $object->$name ?? null;
+            $object->__setInitializer($originalInitializer);
+
+            return $val;
+        }
+
+        return isset($object->$name) ? parent::getValue($object) : null;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Avoids triggering lazy loading via `__set` if the provided object
+     * is a {@see \Doctrine\Common\Proxy\Proxy}.
+     *
+     * @link https://bugs.php.net/bug.php?id=63463
+     */
+    public function setValue($object, $value = null)
+    {
+        if (! ($object instanceof Proxy && ! $object->__isInitialized())) {
+            parent::setValue($object, $value);
+
+            return;
+        }
+
+        $originalInitializer = $object->__getInitializer();
+        $object->__setInitializer(null);
+        parent::setValue($object, $value);
+        $object->__setInitializer($originalInitializer);
+    }
+}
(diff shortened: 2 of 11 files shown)


tests/Doctrine/Tests_PHP74/Persistence/Reflection/TypedNoDefaultReflectionPropertyTest.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $reflection = new TypedNoDefaultReflectionProperty(TypedNullableFoo::class, 'value');
+        $reflection->setAccessible(true);
+
+        $object = new TypedNullableFoo();
+
+        $reflection->setValue($object, null);
+
+        self::assertNull($reflection->getValue($object));
+        self::assertTrue($reflection->isInitialized($object));
```
greg0ire, line 53: This assertion fails and I can't understand why… can anybody help?

composer.json
```diff
@@ -24,12 +24,12 @@
...
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
-        "doctrine/event-manager": "^1.0",
-        "doctrine/reflection": "^1.2"
+        "doctrine/event-manager": "^1.0"
     },
     "require-dev": {
         "phpstan/phpstan": "^0.11",
         "doctrine/coding-standard": "^6.0",
+        "doctrine/common": "^3.0",
```
greg0ire, line 32: So this is a circular dependency… yuck! It was a dev dependency in `doctrine/reflection`, not sure why… should I move it to a regular requirement?

tests/Doctrine/Tests_PHP74/Persistence/Reflection/TypedNoDefaultReflectionPropertyTest.php
greg0ire: @sspat maybe?

greg0ire: Ah I think I undestand what happened: I refactored the source class from my vendor dir, but copied the test from an outside dir since tests are not shipped in archives, and ended up with a version of the test that does not match the class.

composer.json
morozov: Can you check what will fail if this dependency is removed? A dev dependency of one package is not necessarily a dev dependency of the one that depends on it.

greg0ire: I forgot to add it at first and it failed :slightly_smiling_face:

lib/Doctrine/Persistence/Reflection/RuntimePublicReflectionProperty.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace Doctrine\Persistence\Reflection;
+
+use Doctrine\Common\Proxy\Proxy;
```
morozov, line 5: So this is where the package depends on the common library. It doesn't require it but supports it (hence the test). Would it make sense to just make the test skip if the library is not loaded and add it as a suggestion?

greg0ire: So to be clear, you would add `composer require --dev doctrine/common` to the CI script? In the end, the difference it would make would be that contributors would run part of the tests and would not have to download that lib, at the expense of slightly longer builds in our CI, so… I'm not really convinced but can do the change. Let's see if other contributors feel strongly one way or another?

morozov: > So to be clear, you would add `composer require --dev doctrine/common` to the CI script?

Yes, this was the idea. Technically, the development of the script doesn't strictly depend on this package. But I see what you're saying.

greg0ire: > How is it different from this case?

Here, it doesn't depend on extensions. Extensions are more involving to install: you have to be root, composer can't do it for you, and in some cases you might even not have a system package for them. That's why it makes a lot of sense to proceed like this on the DBAL. `doctrine/common` has 2 dependencies: this very lib, and php, so it means no extra dep. It's just not worth it IMO.

SenseException: I can live with `composer require --dev doctrine/common` or adding it to the dev dependencies. I've seen both ways in different projects.

SenseException (approved):

lib/Doctrine/Persistence/Reflection/RuntimePublicReflectionProperty.php
morozov: If I may summarize it, if there's no reason for the change, let's not make it.

beberlei: The other option is to copy the interface here and add a autoload-dev. Since its just a marker interface with no methods, there is no risk of breaking with or without doctrine/common.

beberlei (approved):

lib/Doctrine/Persistence/Reflection/RuntimePublicReflectionProperty.php
greg0ire: I added the requirement under the `require-dev` section, so it's not like there is an impact on the users.

greg0ire:
I should create a branch called 2.1.x from 2.0.x and target it, shouldn't I?

SenseException:
> I should create a branch called 2.1.x from 2.0.x and target it, shouldn't I? Since master is aliased to 3

I guess so. Is there a reason why master is already covering a 3.0 release?

greg0ire:
> I guess so. Is there a reason why master is already covering a 3.0 release?

Yeah, for instance `master` no longer contains the `PersistentObject` class. Full comparison here: https://github.com/doctrine/persistence/compare/2.0.0...master

Issue #24, opened 2017-05-18
Registry::getManagerForClass doesn't work with anonymous classes


Pierstoval:
### Situation:

* PHP 7.1.5
* Using windows (it's an important part)
* Trying to use Symfony validator to validate anonymous class

The validator point is useless, but what's triggered behind is the<br>`$registry->getManagerForClass(get_class($myAnonymousClassInstance));`

### The test

```php
$a = new class {
    public $value;
};

var_dump(get_class($a));

$manager = $this->getDoctrine()->getManagerForClass(get_class($a));
```` 

Output on Windows:

`class@anonymous\x00E:\dev\www\skeleton\src\AppBundle\Controller\DefaultController.php000001982329027F` 

With this, I get an `ORMException`:

```
Unknown Entity namespace alias 'class@anonymousE'.
```

Obviously, because there is the `:` syntax, Doctrine is looking for a specific entity namespace.

---

I also tested this on linux inside a command and it works:

```php
$a = new class {
    public $value;
};

$manager = $this->getContainer()->get('doctrine')->getManagerForClass(get_class($a));

var_dump($class, $manager);
```

Output is correct because `$manager` is `null`:

```
"class@anonymous\x00/mnt/e/dev/www/skeleton/src/AppBundle/Command/TestCommand.php0x7f66b1aa5341"
null
```

### The fix

I think the `AbstractManagerRegistry::getManagerForClass()` should return `null` if the class name starts with `class@anonymous\x00`, whatever the platform, because it is the way PHP 7.1 handles anonymous classes names.

What do you think?

ZhukV:
Related to https://github.com/symfony/symfony/issues/38008

Anonymous classes contain `:` char, and system try to get real class with namespace used.

greg0ire:
Let's have a look at the phpdoc comment for the parameter of `getMetadataFor`:

https://github.com/doctrine/persistence/blob/ebb6c322b4b438e4156523dafca1aedb1342353a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php#L146

You are supposed to pass a class _name_ . _Anonymous_ classes don't have one, that's why they are called anonymous. 
I'm not sure why `get_class` returns a string in that case, when it could throw instead… anyway it's just a string representation of the anonymous class, but it can't be a name because of what I stated above.

The expected behavior would be to throw an exception, which is already what happens, so there is no bug.

[...]

PR #114, opened 2020-05-26
Allow PHP 8

greg0ire:
We want to let other projects experiment and report bugs to us. Failure
is allowed for now and should no longer be after the feature freeze
starts.

PR #125, opened 2020-08-31
#24 Fix try get metadata for anonymous classes

ZhukV:
Related to https://github.com/symfony/symfony/issues/38008

### DIFF
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -152,6 +152,10 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      */
     public function getMetadataFor($className)
     {
+        if (strpos($className, '@anonymous') !== false) {
+            throw MappingException::anonymousClass($className);
+        }
+
         if (isset($this->loadedMetadata[$className])) {
             return $this->loadedMetadata[$className];
         }
--- a/lib/Doctrine/Persistence/Mapping/MappingException.php
+++ b/lib/Doctrine/Persistence/Mapping/MappingException.php
@@ -92,4 +92,9 @@ class MappingException extends Exception
     {
         return new self(sprintf("Class '%s' does not exist", $className));
     }
+
+    public static function anonymousClass(string $className) : self
+    {
+        return new self(sprintf("The class '%s' is anonymous.", $className));
+    }
 }
--- a/tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
+++ b/tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
@@ -10,6 +10,7 @@ use Doctrine\Persistence\Mapping\MappingException;
 use Doctrine\Tests\DoctrineTestCase;
 use PHPUnit_Framework_MockObject_MockObject;
 use stdClass;
+use function get_class;
 
 /**
  * @covers \Doctrine\Persistence\Mapping\AbstractClassMetadataFactory
@@ -32,6 +33,17 @@ class ClassMetadataFactoryTest extends DoctrineTestCase
         $cache = new ArrayCache();
         $this->cmf->setCacheDriver($cache);
         self::assertSame($cache, $this->cmf->getCacheDriver());
+    }
+
+    public function testGetMetadataForAnonymousClass()
+    {
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessage('anonymous');
+
+        $object = new class () extends stdClass {
+        };
+
+        $this->cmf->getMetadataFor(get_class($object));
     }
 
     public function testGetMetadataFor()


malarzm (approved):
Changes LGTM but I wonder if this should be merged to 1.4.x and then up to 2.0.x. I'll defer the decision to people that usually merge stuff here :)

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -152,6 +152,10 @@ public function getAllMetadata()
      */
     public function getMetadataFor($className)
     {
+        if (strpos($className, 'class@anonymous') === 0) {
```
nicolas-grekas, line 155: should be `strpos($className, '@anonymous') !== false` to be forward compatible with PHP 8

greg0ire:
> Changes LGTM but I wonder if this should be merged to 1.4.x and then up to 2.0.x. I'll defer the decision to people that usually merge stuff here :)

It's a bugfix for #114 I think, so it should probably go to 1.3.x IMO

lib/Doctrine/Persistence/Mapping/MappingException.php
```diff
@@ -92,4 +92,14 @@ public static function nonExistingClass($className)
     {
         return new self(sprintf("Class '%s' does not exist", $className));
     }
+
+    /**
+     * @param string $className
+     *
+     * @return MappingException
+     */
+    public static function anonymousClass($className)
```
greg0ire, line 101: We can use type declarations here:

```suggestion
    public static function anonymousClass(string $className): self
```

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -34,6 +36,19 @@ public function testGetCacheDriver()
...
+    {
+        $object = new class () extends stdClass {
+        };
+
+        try {
+            $this->cmf->getMetadataFor(get_class($object));
+            self::fail('must throw MappingException error');
+        } catch (MappingException $error) {
+            self::assertEquals(1, preg_match("/^The class '([^']+)' is anonymous.$/", $error->getMessage()));
+        }
```
greg0ire, line 49: Let's make the test more simple and robust:

```suggestion
        $this->expectException(MappingException::class);
        $this->expectExceptionMessage('anonymous');
        $this->cmf->getMetadataFor(get_class($object));
```

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
ZhukV: Not know how PHP 8.* represent anonymous class.
Added check only for '@anonymous'.

lib/Doctrine/Persistence/Mapping/MappingException.php
ZhukV: Added type hinting. But, this class in other methods don't use type hinting. As result, I don't add previously.

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
ZhukV: Thank. Fixed. I think what `expectExceptionMessage` full match message - `$error->getMessage() === $expectedMessage`, but I see what `ExceptionMessage Constraint` match by `strpos`.

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -34,6 +35,17 @@ public function testGetCacheDriver()
         self::assertSame($cache, $this->cmf->getCacheDriver());
     }
 
+    public function testGetMetadataForAnonymousClass()
```
greg0ire, line 38:
```suggestion
    public function testGetMetadataForAnonymousClass() : void
```

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -34,6 +35,17 @@ public function testGetCacheDriver()
...
+    {
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessage('anonymous');
+
+        $object = new class () extends stdClass {
+        };
+
+        $this->cmf->getMetadataFor(get_class($object));
```
greg0ire, line 46: It's a good practice to always put the exception expectations right before the call that triggers them.

```suggestion
        $object = new class () extends stdClass {
        };

        $this->expectException(MappingException::class);
        $this->expectExceptionMessage('anonymous');
        
        $this->cmf->getMetadataFor(get_class($object));
```

lib/Doctrine/Persistence/Mapping/MappingException.php
greg0ire: That's because they were introduced at a time when we could not use them. We are probably going to migrate those at some point, so let's do it rightaway for new methods, it will mean less work for the migration.

greg0ire:
Please squash your commits and retarget to 1.3.x

ZhukV:
Create new pull request (#126) based on `v1.3.x`

PR #126, opened 2020-09-02
Bail out early on anonymous class 

### LINKED ISSUE #24: Registry::getManagerForClass doesn't work with anonymous classes

ZhukV:
A colon is used later in this class:

https://github.com/doctrine/persistence/blob/ebb6c322b4b438e4156523dafca1aedb1342353a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php#L160-L166

That's in order to handle aliased namespaces in
which case the argument passed to `getMetadataFor` looks like this:

`alias:ModelClassName`

On Windows, this causes an issue because the string representation of
anonymous classes can contain the driver letter (typically `C:\`), which
includes a colon.

For instance, for
`class@anonymous\x00E:\dev\www\skeleton\src\AppBundle\Controller\DefaultController.php000001982329027F`,
`class@anonymous\x00E` will wrongly be interpreted as a namespace alias
and you will get the following message:

> Unknown Entity namespace alias 'class@anonymousE'.

Fixes #24

### DIFF
--- a/tests/Doctrine/Tests/Common/Persistence/Mapping/ClassMetadataFactoryTest.php
+++ b/tests/Doctrine/Tests/Common/Persistence/Mapping/ClassMetadataFactoryTest.php
@@ -10,6 +10,7 @@ use Doctrine\Persistence\Mapping\MappingException;
 use Doctrine\Tests\DoctrineTestCase;
 use PHPUnit_Framework_MockObject_MockObject;
 use stdClass;
+use function get_class;
 
 /**
  * @covers \Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory
@@ -32,6 +33,17 @@ class ClassMetadataFactoryTest extends DoctrineTestCase
         $cache = new ArrayCache();
         $this->cmf->setCacheDriver($cache);
         self::assertSame($cache, $this->cmf->getCacheDriver());
+    }
+
+    public function testGetMetadataForAnonymousClass() : void
+    {
+        $object = new class () extends stdClass {
+        };
+
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessage('anonymous');
+
+        $this->cmf->getMetadataFor(get_class($object));
     }
 
     public function testGetMetadataFor()
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -154,6 +154,10 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      */
     public function getMetadataFor($className)
     {
+        if (strpos($className, '@anonymous') !== false) {
+            throw MappingException::anonymousClass($className);
+        }
+
         if (isset($this->loadedMetadata[$className])) {
             return $this->loadedMetadata[$className];
         }
--- a/lib/Doctrine/Persistence/Mapping/MappingException.php
+++ b/lib/Doctrine/Persistence/Mapping/MappingException.php
@@ -93,6 +93,11 @@ class MappingException extends Exception
     {
         return new self(sprintf("Class '%s' does not exist", $className));
     }
+
+    public static function anonymousClass(string $className) : self
+    {
+        return new self(sprintf("The class '%s' is anonymous.", $className));
+    }
 }
 
 class_exists(\Doctrine\Common\Persistence\Mapping\MappingException::class);


SenseException (approved):

greg0ire (approved):
I tested locally the codecov comment seems wrong.

ostrolucky (changes requested):
I don't understand how this fixes referenced issue. Issue was that exception was thrown instead of null and this patch throws exception?? Requested fix from OP was to return null consistently in windows and linux.

greg0ire:
@ostrolucky the fix is to have a mapping exception be thrown (which must be properly handled at the call site).

greg0ire:
Actually I'm wrong, a MappingException is already thrown, not sure how this is better…

greg0ire:
It's like so since 4f8515a7d88f254fd6d9e177d0baff8baa9a0511

ostrolucky:
Does that mean that class_exists returns true for anonymous class on Linux, but false in Windows? If that's the case it means it's PHP bug that should be fixed there, I don't see what's wrong in doctrine/persistence.

tests/Doctrine/Tests/Common/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -34,6 +35,17 @@ public function testGetCacheDriver()
         self::assertSame($cache, $this->cmf->getCacheDriver());
     }
 
+    public function testGetMetadataForAnonymousClass() : void
+    {
+        $object = new class () extends stdClass {
+        };
+
+        $this->expectException(MappingException::class);
+        $this->expectExceptionMessage('anonymous');
```
greg0ire, line 44: If we drop this line, we can revert all the changes made in `src`, the test will still pass

greg0ire (changes requested):

tests/Doctrine/Tests/Common/Persistence/Mapping/ClassMetadataFactoryTest.php
greg0ire: Oh right that must be because the bug is specific to Windows… as pointed out in https://github.com/doctrine/persistence/pull/126#issuecomment-694439226

ZhukV: No, it's wrong, because PHP generate custom class name for anonymous classes and we must check, what this exception throws by anonymous class. I understand problem, but in this case we must use one of those cases: generate expected message per PHP version or collect all messages and find universal message. For more details, please see #125 (we close this PR, because this problem must be merged to 1.3 and share to next versions).

greg0ire:
@ostrolucky this is the root of the issue as pointed out by OP in https://github.com/symfony/symfony/issues/38008

greg0ire (approved):
I pushed a new commit with the same changes, but an improved message.

ostrolucky:
I don't think this is root issue. Root issue is that MappingException isn't thrown when class alias branch kicks in. This passes for me on both Windows on Linux.

```php
public function testGetMetadataForAnonymousClass() : void
{

    $this->cmf->getMetadataFor('random:string:wasas');
}
```

But will likely not pass when ORM is hooked in. You should realize we will have same issue no matter if this is anonymous class or not, this is user supplied input so bug will surface anytime user supplies string with colons.

Only thing required by https://github.com/symfony/symfony/issues/38008 is that MappingException is thrown. Can't we change ORM to throw mapping exception instead? That would be more correct fix, as doctrine/persistence correctly throws MappingException already, but ORM manages to throw different exception before that happens.

greg0ire:
I think you are right @ostrolucky , since that method receives strings and not objects, it's up to the caller to detect that they are misusing the API by passing the string representation of an anonymous class.

The only thing that should be fixed here IMO is this comment:

https://github.com/doctrine/persistence/blob/ebb6c322b4b438e4156523dafca1aedb1342353a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php#L146

It should be improved to detail valid names, in particular it should mention aliases.

PR #129, opened 2020-10-12
Fix PHP8 build

greg0ire:

PR #131, opened 2020-10-13
Manually merge 2.0.x into 2.1.x

greg0ire:
There was a conflict in `composer.lock` because of #129 that prevented doing that from the UI.
I'm using the draft status so that no one merges this. I will push instead after getting approval.


ostrolucky (approved):

greg0ire:
Looks like `2.1.x` contains code that does not pass with PHPStan 0.12

PR #132, opened 2020-10-17
Migrate all checks to Github Actions

greg0ire:
It is faster, we get more precise feedback right in Github's UI, and we
can reuse common workflows.


greg0ire:
Blocked by #133

greg0ire:
Also, I need to use `composer update`, not `composer install` otherwise I will not be able to test 7.1 and 8.0

greg0ire:
Rebasing on 2.1.x should help, since there is no requirement on `doctrine/reflection` there.

greg0ire:
When this is approved, I will change the settings so this is mergeable and new jobs are required.

ostrolucky (approved):

PR #133, opened 2020-10-17
Upgrade cs

greg0ire:


ostrolucky (approved):

SenseException (approved):
All in all it's okay, but why did you add return types in some cases and not in other cases for test classes?

It also seems that [one Closure](https://github.com/doctrine/persistence/pull/133/files#diff-a914e7a53c337795b19ccdef140a37b8d13aa44a2a7fa0e50e5a7c7b37d1a4beR171) wasn't marked for its missing return type and 'static`.

greg0ire:
>  but why did you add return types in some cases and not in other cases for test classes?

I think I didn't, it was phpcbf

PR #135, opened 2020-10-18
Manually merge 2.1.x into master

greg0ire:


greg0ire:
There were a lot of PHPStan issues to fix, not sure why…

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -148,19 +145,27 @@ public function testWillFailOnFallbackFailureWithNotLoadedMetadata(): void
     public function testWillIgnoreCacheEntriesThatAreNotMetadataInstances(): void
     {
         $cacheDriver = $this->createMock(Cache::class);
-        assert($cacheDriver instanceof Cache || $cacheDriver instanceof MockObject);
```
ostrolucky: I don't like all of these asserts removed, what causes it?

greg0ire: I think it is a benefit of having this of on `master`: https://github.com/doctrine/persistence/blob/3e22cae868ce96a0ca651a7c7c036c7841af6eaf/composer.json#L32

phpstan complains this assertion will always be true because of the extra insight provided by this plugin.

ostrolucky: Well that's annoying. Those assertions are for phpstorm and they verify what's inside docblocks is true.

greg0ire: Weirdly, your comment is marked as outdated and I can't find the assertion in the diff…

ostrolucky: I commented on last commit, you can find it there. I guess it was added in one of previous commits in this PR and then removed by last commit.

greg0ire: Ah it makes sense. Anyway, I was thinking using a phpdoc `/** @var */` comment could be a solution, but it isn't, because `phpcs` requires that we transform that in an assertion. Tell me if you see another option.

SenseException (approved):

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
greg0ire: ping @ostrolucky should I push this?

ostrolucky: you can merge it, but it would be great if you had a time to check and report the issue with assert() upstream

ostrolucky (approved):

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
greg0ire: The question is, which upstream? `doctrine/coding-standard`? Also, don't you think PHPStorm, will eventually catch up and have its own way to figure this out?

ostrolucky: PHPStan or phpstan plugin is upstream here. Our coding standard does not make people to use phpunit plugin. PHPStorm has long, long way to support every syntax that psalm supports. Even then, it doesn't solve diametral issue that assert is used for actually verifying things runtime. Docblock could specify garbage and phpstan would be happy unless triggering type issue. Also why are no complaints here? https://phpstan.org/r/2f928510-d812-44b9-980b-3a7b88a7e517 As I say, there should be more investigation there, but I don't have extra time now.

greg0ire: Apparently it comes from https://github.com/phpstan/phpstan-strict-rules (and the "always true" checks can be disabled with a setting).

PR #139, opened 2020-10-25
Update branch metadata

greg0ire:


.doctrine-project.json
```diff
@@ -11,21 +11,27 @@
             "upcoming": true
         },
         {
-            "name": "2.1",
-            "branchName": "2.1.x",
-            "slug": "2.1",
+            "name": "2.2",
+            "branchName": "2.2.x",
+            "slug": "latest",
```
SenseException: afaik `slug` has to be unique.

SenseException (changes requested):

.doctrine-project.json
greg0ire: Also, it's idiotic to use `latest` twice :stuck_out_tongue: My mistake.

SenseException (approved):

PR #142, opened 2020-11-18
Fix psalm GA

franmomu:
After https://github.com/psalm/psalm-github-actions/pull/17, we need to use a variable to be able to install dev packages before executing psalm.


greg0ire (approved):

greg0ire:
Thanks @franmomu !

PR #144, opened 2020-11-27
Deprecate using doctrine/cache in favour of PSR-6

alcaeus:
This PR deprecates the `getCacheDriver` and `setCacheDriver` methods in the abstract metadata factory in favour of `setCache` and `getCache` which work with PSR-6 cache instances.

To make the transition easier, this PR provides a full backward and forward compatibility layer. When using `setCache` to use a PSR-6 cache, `getCacheDriver` will expose a doctrine/cache instance that wraps the PSR-6 cache. Similarly, when using `setCacheDriver` with a doctrine/cache instance, `getCache` will expose a PSR-6 cache that wraps the original cache.

The new protected `getCacheKey` method is used to avoid hardcoding cache keys in tests and can also be used by extending implementations to work with cache items directly (although they shouldn't).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
         "doctrine/event-manager": "^1.0",
-        "psr/cache": "^1.0",
+        "psr/cache": "^1.0|^2.0|^3.0",
         "symfony/cache": "^4.4|^5.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,9 @@
         "doctrine/annotations": "^1.0",
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
-        "doctrine/event-manager": "^1.0"
+        "doctrine/event-manager": "^1.0",
+        "psr/cache": "^1.0",
+        "symfony/cache": "^4.4|^5.0"
     },
     "require-dev": {
         "composer/package-versions-deprecated": "^1.11",
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -2,17 +2,27 @@
 
 namespace Doctrine\Persistence\Mapping;
 
+use BadMethodCallException;
 use Doctrine\Common\Cache\Cache;
+use Doctrine\Common\Cache\CacheProvider;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Persistence\Proxy;
+use Psr\Cache\CacheItemPoolInterface;
 use ReflectionException;
+use Symfony\Component\Cache\Adapter\DoctrineAdapter;
+use Symfony\Component\Cache\DoctrineProvider;
 
 use function array_reverse;
 use function array_unshift;
 use function explode;
+use function sprintf;
+use function str_replace;
 use function strpos;
 use function strrpos;
 use function substr;
+use function trigger_error;
+
+use const E_USER_DEPRECATED;
 
 /**
  * The ClassMetadataFactory is used to create ClassMetadata objects that contain all the
@@ -28,10 +38,13 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      *
      * @var string
      */
-    protected $cacheSalt = '$CLASSMETADATA';
+    protected $cacheSalt = '__CLASSMETADATA__';
 
     /** @var Cache|null */
     private $cacheDriver;
+
+    /** @var CacheItemPoolInterface|null */
+    private $cache;
 
     /** @var ClassMetadata[] */
     private $loadedMetadata = [];
@@ -45,21 +58,52 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
     /**
      * Sets the cache driver used by the factory to cache ClassMetadata instances.
      *
+     * @deprecated setCacheDriver was deprecated in doctrine/persistence 2.2 and will be removed in 3.0. Use setCache instead
+     *
      * @return void
      */
     public function setCacheDriver(?Cache $cacheDriver = null)
     {
+        @trigger_error(sprintf('%s is deprecated. Use setCache() with a PSR-6 cache instead.', __METHOD__), E_USER_DEPRECATED);
+
         $this->cacheDriver = $cacheDriver;
+
+        if ($cacheDriver === null) {
+            $this->cache = null;
+
+            return;
+        }
+
+        if (! $cacheDriver instanceof CacheProvider) {
+            throw new BadMethodCallException('Cannot convert cache to PSR-6 cache');
+        }
+
+        $this->cache = new DoctrineAdapter($cacheDriver);
     }
 
     /**
      * Gets the cache driver used by the factory to cache ClassMetadata instances.
      *
+     * @deprecated getCacheDriver was deprecated in doctrine/persistence 2.2 and will be removed in 3.0. Use getCache instead
+     *
      * @return Cache|null
      */
     public function getCacheDriver()
     {
+        @trigger_error(sprintf('%s is deprecated. Use getCache() instead.', __METHOD__), E_USER_DEPRECATED);
+
         return $this->cacheDriver;
+    }
+
+    public function setCache(CacheItemPoolInterface $cache): void
+    {
+        $this->cache       = $cache;
+        $this->cacheDriver = new DoctrineProvider($cache);
+    }
+
+    public function getCache(): ?CacheItemPoolInterface
+    {
+        return $this->cache;
     }
 
     /**
@@ -174,19 +218,20 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
         $loadingException = null;
 
         try {
-            if ($this->cacheDriver) {
-                $cached = $this->cacheDriver->fetch($realClassName . $this->cacheSalt);
+            if ($this->cache) {
+                $cached = $this->cache->getItem($this->getCacheKey($realClassName))->get();
                 if ($cached instanceof ClassMetadata) {
                     $this->loadedMetadata[$realClassName] = $cached;
 
                     $this->wakeupReflection($cached, $this->getReflectionService());
                 } else {
                     foreach ($this->loadMetadata($realClassName) as $loadedClassName) {
-                        $this->cacheDriver->save(
-                            $loadedClassName . $this->cacheSalt,
-                            $this->loadedMetadata[$loadedClassName]
-                        );
+                        $item = $this->cache->getItem($this->getCacheKey($loadedClassName));
+                        $item->set($this->loadedMetadata[$loadedClassName]);
+                        $this->cache->saveDeferred($item);
                     }
+
+                    $this->cache->commit();
                 }
             } else {
                 $this->loadMetadata($realClassName);
@@ -398,6 +443,11 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
         }
 
         return $this->reflectionService;
+    }
+
+    protected function getCacheKey(string $realClassName): string
+    {
+        return str_replace('\\', '__', $realClassName) . $this->cacheSalt;
     }
 
     /**
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -12,6 +12,9 @@ use ReflectionException;
 use Symfony\Component\Cache\Adapter\DoctrineAdapter;
 use Symfony\Component\Cache\DoctrineProvider;
 
+use function array_combine;
+use function array_keys;
+use function array_map;
 use function array_reverse;
 use function array_unshift;
 use function explode;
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -24,7 +24,9 @@
         "doctrine/annotations": "^1.0",
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
-        "doctrine/event-manager": "^1.0"
+        "doctrine/event-manager": "^1.0",
+        "psr/cache": "^1.0",
+        "symfony/cache": "^4.4|^5.0"
```
alcaeus, line 29: This is used to provide the forward/compatibility layer. Once we drop that in 3.0, this becomes a dev dependency.

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -400,6 +445,11 @@ public function getReflectionService()
         return $this->reflectionService;
     }
 
+    protected function getCacheKey(string $realClassName): string
+    {
+        return str_replace('\\', '__', $realClassName) . $this->cacheSalt;
```
alcaeus, line 462: This (along with the changed cache salt) will invalidate existing caches. We'll have to check if existing implementations manually mess with cache keys, in which case we'll have to come back and check the impact of this change.

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -141,20 +168,84 @@ public function testWillFailOnFallbackFailureWithNotLoadedMetadata(): void
...
+            ->method('commit');
+
+        $this->cmf->setCache($cacheDriver);
+
+        self::assertSame($metadata, $this->cmf->getMetadataFor(RootEntity::class));
+    }
 
-        $metadata = $this->createMock(ClassMetadata::class);
-        assert($metadata instanceof ClassMetadata);
+    public function testWillNotCacheFallbackMetadata(): void
```
alcaeus, line 207: I duplicated that test, as I realised that using the `onMetadataNotFound` fallback does not cache that metadata. Since I assume this is intended, I haven't made changes and instead opted to codify that behaviour in tests.

alcaeus (commented):
I've left some feedback below.

@greg0ire what settings do I need to pass to update the lock file? Looks like psalm does not like something about this change, but I assume it's due to composer.lock changes, not the cache change itself.

greg0ire:
I guess you will need to fix the conflicts for a new build to be triggered and the SA to be fixed (I merged up 2.1.x  into 2.2.x, it contains a fix for your issue.)

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -45,23 +59,54 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
...
         $this->cacheDriver = $cacheDriver;
+
+        if ($cacheDriver === null) {
+            $this->cache = null;
+
+            return;
+        }
+
+        if (! $cacheDriver instanceof CacheProvider) {
+            throw new BadMethodCallException('Cannot convert cache to PSR-6 cache');
```
greg0ire, line 81: :thinking: this looks like a BC-break

alcaeus: Yeah, it's the one edge case we can't really do anything about. I'm willing to wait for someone to complain...

greg0ire:
My experience with `doctrine/cache` in paid projects is that it has bad framework integration, so I usually create a Symfony cache pool and wrap it in an adapter so that I can use it as a result cache for Doctrine and also see if the cache is used in the profiler. IMO this is the right move because it will mean we will not have to worry about things such as integrating `doctrine/cache` and a framework. Or maintaining that package. This is a great move IMO.

greg0ire:
Not sure why there are this many PHPStan errors, but a lot of them look like they could be fixed by setting up `phpstan/phpstan-phpunit`.

alcaeus:
> Not sure why there are this many PHPStan errors, but a lot of them look like they could be fixed by setting up `phpstan/phpstan-phpunit`.

Not sure where those are coming from. The problem may be that I need to update the lock file due to dependency changes and it installs a newer version than it did before. We should probably document this in our contribution docs.

alcaeus:
- [ ] A list
- [ ] With more things
- [x] And done things

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -45,23 +62,54 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
     /**
      * Sets the cache driver used by the factory to cache ClassMetadata instances.
      *
+     * @deprecated Deprecated in favor of setCache
```
malarzm: Do we want to use same deprecation templates as in ODM? 

> `setCacheDriver` was deprecated in doctrine/persistence 2.2 and will be removed in 3.0. Use `setCache` instead

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
malarzm: Shall we make a note about this in UPGRADE.md?

malarzm (approved):

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -174,19 +222,29 @@ public function getMetadataFor($className)
...
-                            $this->loadedMetadata[$loadedClassName]
-                        );
+                    $loadedMetadata = $this->loadMetadata($realClassName);
+                    $classNames     = array_combine(
+                        array_map([$this, 'getCacheKey'], $loadedMetadata),
+                        $loadedMetadata
+                    );
+
+                    foreach ($this->cache->getItems(array_keys($classNames)) as $item) {
+                        if (! isset($classNames[$item->getKey()])) {
```
nicolas-grekas, line 238: this check is not needed IIUC

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -400,6 +458,16 @@ public function getReflectionService()
         return $this->reflectionService;
     }
 
+    protected function getCacheKey(string $realClassName): string
+    {
+        return str_replace('\\', '__', $realClassName) . $this->cacheSalt;
+    }
+
+    protected function getClassNameFromCacheKey(string $cacheKey): string
```
nicolas-grekas: this is never used, is it worth having it?

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
alcaeus: I'll introduce the deprecation-contracts package in a separate PR and update these messages then 👍

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
alcaeus: Good point. Unresolving this conversation and will follow up.

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
alcaeus: This could happen if (for some weird reason) the cache returned items that weren't requested. In this case we can skip and silently ignore - just being defensive here.

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
alcaeus: That may be a leftover of some different API attempts. I'll remove it.

derrabus (approved):

composer.json
greg0ire: `symfony/cache` has a dev-dependency on `doctrine/cache`, so this means a circular dependency (not for us, but for them when they are in dev mode). I think this might be OK though since it's not meant to last, and since symfony seems to be resorting to `COMPOSER_ROOT_VERSION` in its CI already: https://github.com/symfony/symfony/search?q=COMPOSER_ROOT_VERSION

Still, we might want to check if we don't break their pipeline after releasing this (or merging it? I think they have builds that use untagged versions).

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -139,18 +164,86 @@ public function testWillFailOnFallbackFailureWithNotLoadedMetadata(): void
...
-
-        $cacheDriver->expects(self::once())->method('fetch')->with('Foo$CLASSMETADATA')->willReturn(new stdClass());
+        $key = $this->cmf->getCacheKey(RootEntity::class);
+
+        $metadata = $this->cmf->metadata;
+
+        $item = $this->createMock(CacheItemInterface::class);
+
+        $item
+            ->expects(self::any())
```
greg0ire: This kind of line has no effect and can be dropped I think.
```suggestion
```

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -139,18 +164,86 @@ public function testWillFailOnFallbackFailureWithNotLoadedMetadata(): void
...
+            ->method('getKey')
+            ->willReturn($key);
+        $item
+            ->expects(self::any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -139,18 +164,86 @@ public function testWillFailOnFallbackFailureWithNotLoadedMetadata(): void
...
+            ->method('get')
+            ->willReturn(new stdClass());
+        $item
+            ->expects(self::once())
+            ->method('set')
+            ->with($metadata);
+
+        $cacheDriver = $this->createMock(CacheItemPoolInterface::class);
+        $cacheDriver
+            ->expects(self::any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -139,18 +164,86 @@ public function testWillFailOnFallbackFailureWithNotLoadedMetadata(): void
...
+    public function testWillNotCacheFallbackMetadata(): void
+    {
+        $key = $this->cmf->getCacheKey('Foo');
+
+        $metadata = $this->cmf->metadata;
+
+        $item = $this->createMock(CacheItemInterface::class);
+
+        $item
+            ->expects(self::any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -139,18 +164,86 @@ public function testWillFailOnFallbackFailureWithNotLoadedMetadata(): void
...
+            ->expects(self::never())
+            ->method('commit');
+
+        $this->cmf->setCache($cacheDriver);
 
         $fallbackCallback = $this->getMockBuilder(stdClass::class)->setMethods(['__invoke'])->getMock();
 
-        $fallbackCallback->expects(self::any())->method('__invoke')->willReturn($metadata);
+        $fallbackCallback
+            ->expects(self::any())
```
greg0ire:
```suggestion
```

composer.json
alcaeus: How is there a circular dependency? This is doctrine/persistence, not doctrine/cache.

greg0ire (approved):

composer.json
derrabus: > How is there a circular dependency?

`symfony/symfony` depends on `doctrine/persistence` and replaces `symfony/cache`. Anyway, we should be good here. 😃

composer.json
```diff
@@ -24,7 +24,9 @@
         "doctrine/annotations": "^1.0",
         "doctrine/cache": "^1.0",
         "doctrine/collections": "^1.0",
-        "doctrine/event-manager": "^1.0"
+        "doctrine/event-manager": "^1.0",
+        "psr/cache": "^1.0",
```
derrabus:
```suggestion
        "psr/cache": "^1.0|^2.0|^3.0",
```

alcaeus:
@greg0ire any idea why psalm suddenly complains about `AbstractClassMetadataFactory::getMetadataFor`? Nothing really changed in between, so I suspect that this is a psalm-related issue 🤔

greg0ire:
Woops sorry I'm only noticing your ping now, and I think you are right, see https://github.com/doctrine/persistence/pull/156

greg0ire:
Ah I need to merge up

composer.json
greg0ire: Right, nevermind my comment, not sure what I meant but I must have been confused.

greg0ire:
Thanks @alcaeus !

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -45,23 +61,54 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
...
     {
+        @trigger_error(sprintf('%s is deprecated. Use getCache() instead.', __METHOD__), E_USER_DEPRECATED);
+
         return $this->cacheDriver;
     }
 
+    public function setCache(CacheItemPoolInterface $cache): void
+    {
+        $this->cache       = $cache;
+        $this->cacheDriver = new DoctrineProvider($cache);
```
stof, line 104: this should use the doctrine/cache PSR-6 provider instead, which is available in the latest version

alcaeus: The `DoctrineAdapter` could be replaced. I've created https://github.com/doctrine/cache/pull/366 to also include this class in doctrine/cache so we no longer have external dependencies for the forward and backward compatibility layers. I'll follow up with a PR here once this is done.

PR #151, opened 2020-12-19
Upgrade to Psalm 4.3.1 and use ramsey/composer-install

franmomu:
Like https://github.com/doctrine/.github/pull/16


SenseException (approved):
Not 100% sure if removing composer.lock in a patch version is fine.

malarzm:
> Not 100% sure if removing composer.lock in a patch version is fine.

As long as our tests are working ok we'll be fine. The presence (or absence) is not important for others :)

greg0ire:
Thanks @franmomu !

PR #152, opened 2020-12-19
Drop support for PHP 7.1

franmomu:
I've seen in other `doctrine` packages that support for `7.1` has been dropped, even `7.2`. I don't know the policy for dropping PHP versions.

I think the order in which the call to `loadMetadataForClass` and `isTransient` is made does not matter in the test I've changed, but maybe I'm wrong.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -15,7 +15,6 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.1"
           - "7.2"
           - "7.3"
           - "7.4"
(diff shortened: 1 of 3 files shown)


greg0ire:
> I've seen in other doctrine packages that support for 7.1 has been dropped, even 7.2. I don't know the policy for dropping PHP versions.

True, but also: https://github.com/doctrine/dbal/pull/4386

franmomu:
> > I've seen in other doctrine packages that support for 7.1 has been dropped, even 7.2. I don't know the policy for dropping PHP versions.
> 
> True, but also: [doctrine/dbal#4386](https://github.com/doctrine/dbal/pull/4386)

oh, I didn't see that one, perfect.

PR #153, opened 2020-12-19
Analyze tests directory with Psalm

franmomu:
The `TestFileDriver::loadMappingFile()` method now comply with the interface contract returning `ClassMetadata` instances.

EDIT: This would allow to add `class-string` to type declarations.

### DIFF
--- a/tests/Doctrine/Tests/Persistence/Mapping/SymfonyFileLocatorTest.php
+++ b/tests/Doctrine/Tests/Persistence/Mapping/SymfonyFileLocatorTest.php
@@ -84,6 +84,7 @@ class SymfonyFileLocatorTest extends DoctrineTestCase
         $path   = __DIR__ . '/_files';
         $prefix = 'Foo';
 
+        /** @psalm-suppress NullArgument */
         new SymfonyFileLocator([$path => $prefix], '.yml', null);
     }
 
--- a/tests/Doctrine/Tests/Persistence/Mapping/SymfonyFileLocatorTest.php
+++ b/tests/Doctrine/Tests/Persistence/Mapping/SymfonyFileLocatorTest.php
@@ -84,7 +84,6 @@ class SymfonyFileLocatorTest extends DoctrineTestCase
         $path   = __DIR__ . '/_files';
         $prefix = 'Foo';
 
-        /** @psalm-suppress NullArgument */
         new SymfonyFileLocator([$path => $prefix], '.yml', null);
     }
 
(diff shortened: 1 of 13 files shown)


tests/Doctrine/Tests/Persistence/Mapping/SymfonyFileLocatorTest.php
```diff
@@ -84,6 +84,7 @@ public function testInvalidCustomNamespaceSeparator(): void
         $path   = __DIR__ . '/_files';
         $prefix = 'Foo';
 
+        /** @psalm-suppress NullArgument */
```
SenseException: Isn't this covered by the psalm config?

franmomu: It's covered in phpstan configuration. 

I could add it also to psalm config instead of this annotation here, I usually use psalm annotations because once it's fixed, psalm doesn't complain about ignoring an unexisting error, so I find it easy to remove it from here, but I'll move it if you want.

greg0ire:
Don't mind me, just restarting the build to get a list of jobs that should pass for usage in the settings.

tests/Doctrine/Tests/Persistence/Mapping/SymfonyFileLocatorTest.php
greg0ire: Please move it, we like Psalm to complain about unused ignored rules, IMO it's less confusing for people who know something has been fixed in Psalm but still see the ignore rule.

franmomu: Done, what I meant is that Psalm does not complain about unused ignore rules.

psalm-baseline.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="3.18.2@19aa905f7c3c7350569999a93c40ae91ae4e1626">
```
greg0ire: What's the point of moving Psalm issues to a baseline file? We will have to update this line every time we upgrade Psalm, right? If yes, how?

franmomu: It's true that I think it's usually used when having a big baseline, when you don't want do deal with issues from the start, so I've moved the issue to `psalm.xml`. Out of curiosity, for updating the baseline: `vendor/bin/psalm --update-baseline`.

greg0ire (approved):

greg0ire:
Thanks @franmomu !

franmomu:
Thanks @greg0ire, can you please merge up with `2.2.x` whenever you can, I wanted to try to add generics to `ClassMetadata`.

greg0ire:
Done, you can go ahead mate :)

PR #155, opened 2021-02-05
I've added missing phpDoc for flush method in ObjectManager.php

foxybytespl:
flush () method can throw UniqueConstraintViolationException and that information was missing in documentation block so I added @throws statement this
will prevent IDE from marking code with warning.


malarzm:
Unfortunately `UniqueConstraintViolationException` is not part of doctrine/persistence (it is DBAL's exception) therefore it can't be declared here. Thanks for taking your time trying to improve our project!

PR #159, opened 2021-02-14
Raise Psalm level to 3 and add generics to ClassMetadata

franmomu:
The first commits of this PR are changes in order to raise psalm level to 3. I've used `2.2.x` as base thinking about adding the `ClassMetadata` generics as feature, but maybe I should split the PR in two targeting `2.1.x` with the raising psalm level part.

### DIFF
--- a/lib/Doctrine/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Persistence/AbstractManagerRegistry.php
@@ -29,7 +29,10 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
     /** @var string */
     private $defaultManager;
 
-    /** @var string */
+    /**
+     * @var string
+     * @psalm-var class-string
+     */
     private $proxyInterfaceName;
 
     /**
@@ -39,6 +42,8 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
      * @param string   $defaultConnection
      * @param string   $defaultManager
      * @param string   $proxyInterfaceName
+     *
+     * @psalm-param class-string $proxyInterfaceName
      */
     public function __construct($name, array $connections, array $managers, $defaultConnection, $defaultManager, $proxyInterfaceName)
     {
--- a/lib/Doctrine/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Persistence/AbstractManagerRegistry.php
@@ -163,13 +163,9 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
      */
     public function getManagerForClass($class)
     {
-        // Check for namespace alias
-        if (strpos($class, ':') !== false) {
-            [$namespaceAlias, $simpleClassName] = explode(':', $class, 2);
-            $class                              = $this->getAliasNamespace($namespaceAlias) . '\\' . $simpleClassName;
-        }
+        $className = $this->getRealClassName($class);
 
-        $proxyClass = new ReflectionClass($class);
+        $proxyClass = new ReflectionClass($className);
 
         if ($proxyClass->implementsInterface($this->proxyInterfaceName)) {
             $parentClass = $proxyClass->getParentClass();
@@ -178,18 +174,37 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
                 return null;
             }
 
-            $class = $parentClass->getName();
+            $className = $parentClass->getName();
         }
 
         foreach ($this->managers as $id) {
             $manager = $this->getService($id);
 
-            if (! $manager->getMetadataFactory()->isTransient($class)) {
+            if (! $manager->getMetadataFactory()->isTransient($className)) {
                 return $manager;
             }
         }
 
         return null;
+    }
+
+    /**
+     * @param string $className
+     *
+     * @psalm-return class-string
+     */
+    private function getRealClassName($className): string
+    {
+        // Check for namespace alias
+        if (strpos($className, ':') !== false) {
+            [$namespaceAlias, $simpleClassName] = explode(':', $className, 2);
+
+            /** @psalm-var class-string */
+            return $this->getAliasNamespace($namespaceAlias) . '\\' . $simpleClassName;
+        }
+
+        /** @psalm-var class-string */
+        return $className;
     }
 
     /**
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -163,6 +163,8 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      * @param string $simpleClassName
      *
      * @return string
+     *
+     * @psalm-return class-string
      */
     abstract protected function getFqcnFromAlias($namespaceAlias, $simpleClassName);
 
@@ -307,6 +309,8 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      * @param string $name
      *
      * @return string[]
+     *
+     * @psalm-param class-string $name
      */
     protected function getParentClasses($name)
     {
@@ -337,6 +341,8 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      * @param string $name The name of the class for which the metadata should get loaded.
      *
      * @return string[]
+     *
+     * @psalm-param class-string $name
      */
     protected function loadMetadata($name)
     {
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -208,6 +208,8 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      *
      * @throws ReflectionException
      * @throws MappingException
+     *
+     * @psalm-param class-string|string $className
      */
     public function getMetadataFor($className)
     {
@@ -433,6 +435,8 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-param class-string|string $class
      */
     public function isTransient($class)
     {
@@ -446,6 +450,7 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
             $class                              = $this->getFqcnFromAlias($namespaceAlias, $simpleClassName);
         }
 
+        /** @psalm-var class-string $class */
         return $this->getDriver()->isTransient($class);
     }
 
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -17,6 +17,7 @@ use function array_keys;
 use function array_map;
 use function array_reverse;
 use function array_unshift;
+use function assert;
 use function explode;
 use function sprintf;
 use function str_replace;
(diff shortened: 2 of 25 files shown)


lib/Doctrine/Persistence/Reflection/TypedNoDefaultReflectionProperty.php
```diff
@@ -38,7 +38,10 @@ public function setValue($object, $value = null)
                 unset($this->$propertyName);
             };
             $unsetter = $unsetter->bindTo($object, $this->getDeclaringClass()->getName());
-            $unsetter();
+
+            if ($unsetter !== false) {
+                $unsetter();
+            }
```
greg0ire: I think an assertion should be used here, instead of silently not calling the `$unsetter` when `bindTo` fails for some reason.

greg0ire:
I don't have a strong opinion about the target branch. Do what you feel is best.

greg0ire (approved):

greg0ire:
Please run `phpcbf`

franmomu:
> Please run `phpcbf`

oops, didn't see that, done.

greg0ire:
@orklah please review

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -218,6 +221,7 @@ public function getMetadataFor($className)
 
             $realClassName = $this->getFqcnFromAlias($namespaceAlias, $simpleClassName);
         } else {
+            /** @psalm-var class-string $className */
```
orklah, line 226: This should be specified in parameters instead of overrided with @var

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -483,6 +491,8 @@ private function getRealClass(string $class): string
             $this->createDefaultProxyClassNameResolver();
         }
 
+        assert($this->proxyClassNameResolver instanceof ProxyClassNameResolver);
```
orklah: asserting the variable is !== null would be semantically more meaningful I think

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -66,7 +66,7 @@ abstract class AnnotationDriver implements MappingDriver
     /**
      * Name of the entity annotations as keys.
      *
-     * @var string[]
+     * @var array<class-string, mixed>
```
orklah: was `string[]` plain wrong? If not, it should become `array<class-string, string>`

lib/Doctrine/Persistence/Mapping/ReflectionService.php
```diff
@@ -21,6 +21,9 @@ interface ReflectionService
      * @return string[]
      *
      * @throws MappingException
+     *
+     * @psalm-param class-string $class
+     * @psalm-return class-string[]
```
orklah, line 26: maybe use the extensive syntax if offset type is known?

psalm.xml
```diff
@@ -39,5 +39,21 @@
                 <file name="tests/Doctrine/Tests/Persistence/Mapping/SymfonyFileLocatorTest.php"/>
             </errorLevel>
         </NullArgument>
+        <PossiblyNullReference>
```
orklah, line 42: Maybe leave a comment to explain why those couldn't be fixed and had to be ignored? It will be helpful for future contributions

tests/Doctrine/Tests/Persistence/Mapping/AnnotationDriverTest.php
```diff
@@ -24,7 +24,7 @@ public function testGetAllClassNames(): void
 
 class SimpleAnnotationDriver extends AnnotationDriver
 {
-    /** @var bool[] */
+    /** @var array<class-string, mixed> */
```
orklah: should probably be `array<class-string, bool>`

tests/Doctrine/Tests/Persistence/Mapping/DriverChainTest.php
```diff
@@ -84,17 +85,19 @@ public function testIsTransient(): void
...
      */
     public function testDefaultDriver(): void
     {
-        $companyDriver    = $this->createMock(MappingDriver::class);
-        $defaultDriver    = $this->createMock(MappingDriver::class);
-        $entityClassName  = 'Doctrine\Tests\ORM\Mapping\DriverChainEntity';
+        $companyDriver = $this->createMock(MappingDriver::class);
+        $defaultDriver = $this->createMock(MappingDriver::class);
+        /** @psalm-var class-string */
+        $entityClassName = 'Doctrine\Tests\ORM\Mapping\DriverChainEntity';
```
orklah: I think ::class should be used too (and eventually suppressed because the dependancy is not there) instead of being overrided

tests/Doctrine/Tests/Persistence/Mapping/TestClassMetadataFactory.php
```diff
@@ -91,6 +92,8 @@ protected function onNotFoundMetadata($className)
 
     /**
      * {@inheritDoc}
+     *
+     * @param string $class
```
orklah: I think I remember @inheritDoc have weird behaviour when a part is overrided. Maybe ditch it completely?

tests/Doctrine/Tests/Persistence/ObjectManagerDecoratorTest.php
```diff
@@ -22,7 +22,7 @@ public function __construct(ObjectManager $wrapped)
 
 class ObjectManagerDecoratorTest extends TestCase
 {
-    /** @var MockObject|ObjectManager */
+    /** @var MockObject&ObjectManager */
```
orklah, line 25: Not sure how IDE will react to that. Maybe in an @psalm-var?

lib/Doctrine/Persistence/AbstractManagerRegistry.php
```diff
@@ -173,20 +174,39 @@ public function getManagerForClass($class)
...
+        // Check for namespace alias
+        if (strpos($className, ':') !== false) {
+            [$namespaceAlias, $simpleClassName] = explode(':', $className, 2);
+
+            /** @psalm-var class-string */
+            return $this->getAliasNamespace($namespaceAlias) . '\\' . $simpleClassName;
+        }
+
+        /** @psalm-var class-string */
+        return $className;
```
orklah, line 207: Maybe an assertion for class_exists could be used in the future instead of overriding? In a future version to avoid BC breaks maybe?

orklah (commented):
Great job :)

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
franmomu: Thank you for reviewing! I remember struggling with this one:

```php
// Check for namespace alias
if (strpos($className, ':') !== false) {
    [$namespaceAlias, $simpleClassName] = explode(':', $className, 2);

    $realClassName = $this->getFqcnFromAlias($namespaceAlias, $simpleClassName);
} else {
    /** @psalm-var class-string $className */
    $realClassName = $this->getRealClass($className);
}
```

`$className` can be a `class-string` or a class name `string` alias (`App:Entity`), I tried to move it to parameters as  `class-string|string` but then `psalm` doesn't know that which one is receiving.

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
franmomu: Looks like it was wrong, I used `mixed` because looking at its use:
- `doctrine/orm` the values are `int`: https://github.com/doctrine/orm/blob/f8fa0fe0694ee661bbcc18406475fdb1749b0be1/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L52-L55
- `doctrine/couchdb-odm` are `bool` : https://github.com/doctrine/couchdb-odm/blob/9bf3d8ed9a649d91b6a66aa439b05a8a1f1ef7fd/lib/Doctrine/ODM/CouchDB/Mapping/Driver/AnnotationDriver.php#L24-L28

And here in tests we're using `bool`, I'll change it to `array<class-string, int|bool>`.

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
orklah: nevermind then, I thought $className would always be a className. How naive :D

lib/Doctrine/Persistence/Mapping/ReflectionService.php
franmomu: The only implementation I've seen (apart from the ones returning empty array) is

https://github.com/doctrine/persistence/blob/d8de1720d6f526f90042d11a7c74773a0524bb3a/lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php#L34-L41

So I don't know if that can be improved.

tests/Doctrine/Tests/Persistence/ObjectManagerDecoratorTest.php
franmomu: Interesting, I've been using it for a while because `PHPStorm` understands it, don't know others.

orklah: IIRC, PHPStorm basically treat that the same way as "|" now. There was an issue on their bug tracker where they talked about that and they were not sure of what to do with that

tests/Doctrine/Tests/Persistence/Mapping/DriverChainTest.php
```diff
@@ -118,7 +118,7 @@ public function testDefaultDriver(): void
         self::assertNull($chain->getDefaultDriver());
 
         $chain->setDefaultDriver($defaultDriver);
-        $chain->addDriver($companyDriver, 'Doctrine\Tests\Models\Company');
+        $chain->addDriver($companyDriver, 'Doctrine\Tests\Persistence\Mapping\Fixtures\Manager');
```
orklah, line 121: This should be a ::class instead

psalm.xml
```diff
@@ -41,17 +41,21 @@
         </NullArgument>
         <PossiblyNullReference>
             <errorLevel type="suppress">
+                <!-- ReflectionProperty::getType() can return null, but there is ReflectionProperty::hasType()
```
orklah, line 44: This has been fixed in Psalm recently:
https://github.com/vimeo/psalm/issues/5258
Maybe Psalm can be upgraded?

franmomu: Here is `ReflectionProperty` (instead of `ReflectionParameter`) I've created https://github.com/vimeo/psalm/pull/5584 based on https://github.com/vimeo/psalm/issues/5258, but after reading it again, IIUC, it won't work because `ReflectionProperty` is not immutable.

tests/Doctrine/Tests/Persistence/Mapping/DriverChainTest.php
franmomu: In this case, this argument is a namespace.

psalm.xml
orklah: Oh,  my mistake then :)

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
vv12131415: maybe rename `$className` to `$classNameOrClassNameAlias`?
This will ease reading the code then

greg0ire:
This can be closed, right?

PR #160, opened 2021-03-10
[Reflection] Use default typed property value when setting value to null

ruudk:
Related to https://github.com/doctrine/orm/issues/7999

I'm using typed properties with a default value that is not null. Example:
```php
class Entity {
    /**
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private int $id = 0;

    public function getId() : int
    {
        return $this->id;
    }
}
```

This works great. When inserting a new entity in the database, the auto incremented id is updated properly.

It means that entities that are not persisted will have a `0` as id and the rest will have another non-zero integer value.

But upon deleting an entity this becomes a problem. The ORM tries to set the `id` property value to `null` and that causes a TypeError:
```
TypeError: Typed property Entity::$id must be int, null used

vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:1246
vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:441
vendor/doctrine/orm/lib/Doctrine/ORM/EntityManager.php:376
```

With this change applied, the default value of the `id` property `0` is used again.

Solving the problem.

I'm wondering: Does this make any sense? Is this crazy? Am I overlooking something, or is this the solution? 😊

### DIFF
--- a/lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
+++ b/lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
@@ -4,6 +4,7 @@ namespace Doctrine\Persistence\Mapping;
 
 use Doctrine\Persistence\Reflection\RuntimePublicReflectionProperty;
 use Doctrine\Persistence\Reflection\TypedNoDefaultReflectionProperty;
+use Doctrine\Persistence\Reflection\TypedWithDefaultReflectionProperty;
 use ReflectionClass;
 use ReflectionException;
 use ReflectionMethod;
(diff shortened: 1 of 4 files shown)


SenseException:
Thanks for contributing your pull request. I'm not sure about the use case. What is the reason that you prefer to use a 0 valued id instead of a nullable type?

stof:
@SenseException I would say, to avoid making `getId()` return a nullable type when any persisted entity would never have `null`

SenseException:
In case this is about a check if the entity has an id, `0 !== $entity->getId()` can be handled internally in the entity itself with `$entity->hasId()` or something similar.

ruudk:
Closing as I lost interesting in solving it this way.... Thanks for the feedback though.

PR #161, opened 2021-03-15
Add `$orderBy` to ObjectRepository->findOneBy()

JesusValeraDev:
It doesn't make sense the interface has this signature:
```php
interface ObjectRepository
{
    // . . .

    /**
     * Finds a single object by a set of criteria.
     *
     * @param mixed[] $criteria The criteria.
     *
     * @return object|null The object.
     *
     * @psalm-return T|null
     */
    public function findOneBy(array $criteria);
}
```

And the implementation has this one:
```php
class EntityRepository implements ObjectRepository, Selectable
{
    // . . .

    /**
     * Finds a single entity by a set of criteria.
     *
     * @param array      $criteria
     * @param array|null $orderBy
     *
     * @return object|null The entity instance or NULL if the entity can not be found.
     *
     * @psalm-return ?T
     */
    public function findOneBy(array $criteria, array $orderBy = null)
    {
        $persister = $this->_em->getUnitOfWork()->getEntityPersister($this->_entityName);

        return $persister->load($criteria, null, null, [], null, 1, $orderBy);
    }
}
```

### DIFF
--- a/lib/Doctrine/Persistence/ObjectRepository.php
+++ b/lib/Doctrine/Persistence/ObjectRepository.php
@@ -55,12 +55,13 @@ interface ObjectRepository
      * Finds a single object by a set of criteria.
      *
      * @param mixed[] $criteria The criteria.
+     * @param mixed[]|null $orderBy The order by.
      *
      * @return object|null The object.
      *
      * @psalm-return T|null
      */
-    public function findOneBy(array $criteria);
+    public function findOneBy(array $criteria, array $orderBy = null);
 
     /**
      * Returns the class name of the object managed by the repository.


lib/Doctrine/Persistence/ObjectRepository.php
```diff
@@ -55,12 +55,13 @@ public function findBy(array $criteria, ?array $orderBy = null, $limit = null, $
...
      *
      * @param mixed[] $criteria The criteria.
+     * @param mixed[]|null $orderBy The order by.
      *
      * @return object|null The object.
      *
      * @psalm-return T|null
      */
-    public function findOneBy(array $criteria);
+    public function findOneBy(array $criteria, array $orderBy = null);
```
greg0ire, line 64: This is a BC-break for all implementations which means you should target 3.0.x. Besides, I don't understand your comment, why wouldn't it make sense to have have an implementation supporting more calls than what is described in this interface? If they want to have extra features, they should be able to.

JesusValeraDev: Hello @greg0ire 

Yes, this would be for the 3.0.x in case of approvement, of course. 😃 
The `$orderBy` parameter is optional to avoid BC.

The idea of this PR is to avoid having different signatures between the interface and the implementation.
I find it really useful in stricter languages like Java where this is not possible.

The interfaces are used to invert the dependencies between the interface and the implementation.

Here is a picture trying to use the findOneBy method from the `EntityRepository` but using the `ObjectRepository` interface. Because when doing this:
```php
private ObjectRepository $repository;

public function __construct(EntityManagerInterface $entityManager)
{
    $this->repository = $entityManager->getRepository(Order::class);
}
```
The `$repository` object is not an `OrderRepository`(which doesn't exist)  nor `EntityRepository` but `ObjectRepository`.

The `option 1` is marked with a warning, you need to find a workaround like `option 2` to remove it, but also, in this case, this is not exactly what we want to do (initially at `findBy()`, we are fetching more than one value).

image

stof: Adding a parameter (optional or not) in an interface is a BC break for packages **implementing** the interface.

and all implementers of the interface are in other packages...

JesusValeraDev: Ok, I understand.
we can close the issue then.

Thank you 😄

PR #164, opened 2021-04-02
Define `Proxy` as generic

franmomu:
This would allow things like https://github.com/doctrine/common/pull/926

### DIFF
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -472,6 +472,10 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
 
     /**
      * Gets the real class name of a class name that could be a proxy.
+     *
+     * @template T of object
+     * @psalm-param class-string<Proxy<T>>|class-string<T> $class
+     * @psalm-return class-string<T>
      */
     private function getRealClass(string $class): string
     {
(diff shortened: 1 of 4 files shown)


greg0ire (approved):

greg0ire:
@franmomu maybe this can be improved?

https://github.com/doctrine/persistence/blob/12a05f8d1ca070f491b25a82ba1b8e9448da9f21/lib/Doctrine/Persistence/Mapping/ProxyClassNameResolver.php#L8-L9

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -485,14 +489,21 @@ private function getRealClass(string $class): string
     private function createDefaultProxyClassNameResolver(): void
     {
         $this->proxyClassNameResolver = new class implements ProxyClassNameResolver {
+            /**
+             * @template T of object
+             * @psalm-param class-string<Proxy<T>>|class-string<T> $className
```
franmomu, line 494: I've copied this because otherwise the `/** @psalm-var class-string<T> */` won't recognise the `T`.

VincentLanglet: Seems like this is not enough, the tests are still failing

franmomu: Apparently [it should work](https://github.com/phpstan/phpstan/discussions/4802#discussioncomment-567960), I'll ignore for now.

VincentLanglet (approved):

greg0ire:
Thanks @franmomu !

PR #167, opened 2021-04-19
Generics for cmf

greg0ire:
This should make a lot of static analysis issues vanish, because tools
will be able to understand that they are guaranteed to receive some
specialization of ClassMetadata, or that they should provide it.

Also upgraded to doctrine/coding-standard 9 so make sure to review commit by commit, it will be easier.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,11 +30,11 @@
     },
     "require-dev": {
         "composer/package-versions-deprecated": "^1.11",
-        "phpstan/phpstan": "^0.12.84",
+        "phpstan/phpstan": "0.12.84",
         "doctrine/coding-standard": "^6.0 || ^9.0",
         "doctrine/common": "^3.0",
         "phpunit/phpunit": "^7.5.20 || ^8.0 || ^9.0",
-        "vimeo/psalm": "^4.3.1"
+        "vimeo/psalm": "4.7.0"
     },
     "conflict": {
         "doctrine/common": "<2.10@dev"
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     },
     "require-dev": {
         "composer/package-versions-deprecated": "^1.11",
-        "phpstan/phpstan": "^0.12",
+        "phpstan/phpstan": "^0.12.84",
         "doctrine/coding-standard": "^6.0 || ^9.0",
         "doctrine/common": "^3.0",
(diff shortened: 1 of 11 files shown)


composer.json
```diff
@@ -31,7 +31,7 @@
     "require-dev": {
         "composer/package-versions-deprecated": "^1.11",
         "phpstan/phpstan": "^0.12",
```
stof: shouldn't this be bumped to `^0.12.84` as you rely on the fixed behavior of that release ?

greg0ire: I didn't follow the latest developments on PHPStan so I don't know that I am, but sure, let's do this 👍

stof: well, look at the diff on the baseline file :smile: 
The removed entry had a link to a phpstan issue, which got solved in 0.12.84. That's how I found out the necessary version

greg0ire: Oh, right, already forgot about that :sweat_smile:

greg0ire:
@orklah please review

beberlei (approved):

simPod (approved):
I did not checkout the code and don't have thorough overview over doctrine/persistence but the changes I see on Github seem 👌🏾

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -232,6 +238,7 @@ public function getMetadataFor($className)
             if ($this->cache) {
                 $cached = $this->cache->getItem($this->getCacheKey($realClassName))->get();
                 if ($cached instanceof ClassMetadata) {
+                    /** @psalm-var CMTemplate $cached */
```
orklah, line 241: This feels like cheating, but without trying to wrap the cache interface, I'm not sure we can do better

composer.json
```diff
@@ -30,8 +30,8 @@
     },
     "require-dev": {
         "composer/package-versions-deprecated": "^1.11",
-        "phpstan/phpstan": "^0.12",
-        "doctrine/coding-standard": "^6.0 || ^8.0",
+        "phpstan/phpstan": "^0.12.84",
```
orklah: I'd advise pinning the version of static tools, especially now that we started deploying baseline. It will make sure the version used to generate the baseline is the same for everyone, and make sure not to break builds the day before the release :)

orklah (approved):

PR #169, opened 2021-04-21
Use psalm errorLevel 3

franmomu:
Another part from https://github.com/doctrine/persistence/pull/159

### DIFF
--- a/lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
+++ b/lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
@@ -61,7 +61,10 @@ class RuntimeReflectionService implements ReflectionService
     }
 
     /**
-     * {@inheritDoc}
+     * @param string $class
+     * @psalm-param class-string $class
+     *
+     * @return ReflectionClass
      */
     public function getClass($class)
     {
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,8 +1,7 @@
 <?xml version="1.0"?>
 <psalm
-    totallyTyped="false"
-    errorLevel="4"
-    resolveFromConfigFile="true"
+    errorLevel="3"
+    findUnusedPsalmSuppress="true"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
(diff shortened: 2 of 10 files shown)


tests/Doctrine/Tests/Persistence/ObjectManagerDecoratorTest.php
```diff
@@ -22,7 +22,7 @@ public function __construct(ObjectManager $wrapped)
 
 class ObjectManagerDecoratorTest extends TestCase
 {
-    /** @var MockObject|ObjectManager */
+    /** @var MockObject&ObjectManager */
```
franmomu, line 25: About this, it was included in PHPStorm apparently in 2018:https://blog.jetbrains.com/phpstorm/2018/09/phpstorm-2018-3-eap-183-2635-12/

Let me now if we should use `@psalm-var`.

stof: PHPStorm indeed added parsing for it in 2.018 (it was still _treating_ it as a union type internally as it did not support intersection types, but that's not a reason to avoid it in favor of writing a union type directly).

So +1 from me

psalm.xml
```diff
@@ -39,5 +39,24 @@
                 <file name="tests/Doctrine/Tests/Persistence/Mapping/SymfonyFileLocatorTest.php"/>
             </errorLevel>
         </NullArgument>
+        <PossiblyNullReference>
+            <errorLevel type="suppress">
+                <!-- ReflectionProperty::getType() can return null, but there is ReflectionProperty::hasType()
+                check before -->
+                <file name="lib/Doctrine/Persistence/Reflection/TypedNoDefaultReflectionProperty.php"/>
+            </errorLevel>
+        </PossiblyNullReference>
```
franmomu: For things like this that can be fixed in next psalm releases, can we maybe use annotations (`@psalm-suppress`) so we can enable [findUnusedPsalmSuppress](https://psalm.dev/docs/running_psalm/configuration/#findunusedpsalmsuppress) and if this is fixed, psalm would warn us.

greg0ire: Yes please!

psalm.xml
```diff
@@ -1,8 +1,7 @@
 <?xml version="1.0"?>
 <psalm
-    totallyTyped="false"
```
franmomu, line 3: [totallyTyped](https://psalm.dev/docs/running_psalm/configuration/#totallytyped) is deprecated and [`resolveFromConfigFile` by default is `true`](https://github.com/vimeo/psalm/blob/0d524ecd5b4bea45de1aae77dd1638d4ea03457d/config.xsd#L62).

psalm.xml
franmomu: Hmm I added this `PossiblyNullReference` in the first place because locally I used `vendor/bin/psalm --no-cache --php-version=7.4` and running that command it throws the `PossiblyNullReference`.

But since we are not specifying the php version here, I think it's using the lowest one from `composer.json`, `7.1` and using that one there is no error.

greg0ire (approved):

greg0ire:
@orklah please review :slightly_smiling_face:

SenseException (approved):

lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
```diff
@@ -62,6 +62,8 @@ public function getClassNamespace($class)
 
     /**
      * {@inheritDoc}
```
orklah: Are you sure this works properly? I have a vague memory if reading that inheritDoc was ignored by Psalm when there was other tags in the block

orklah (approved):

lib/Doctrine/Persistence/Mapping/RuntimeReflectionService.php
franmomu: You're right, I thought it read both since I got some other error about instantiating `ReflectionClass`, I've added the other tags.

greg0ire:
Thanks @franmomu !

PR #170, opened 2021-04-21
Remove use of deprecated TestCase::at() method

franmomu:
For that test it does not add value to check the order of calls.

https://github.com/doctrine/persistence/blob/fdef2dd7da053bc97841fb45b866a7756bfe0f2f/tests/Doctrine/Tests/Persistence/Mapping/DriverChainTest.php#L26-L40

That order is because there is a call first to:

https://github.com/doctrine/persistence/blob/fdef2dd7da053bc97841fb45b866a7756bfe0f2f/tests/Doctrine/Tests/Persistence/Mapping/DriverChainTest.php#L38

and then to:

https://github.com/doctrine/persistence/blob/fdef2dd7da053bc97841fb45b866a7756bfe0f2f/tests/Doctrine/Tests/Persistence/Mapping/DriverChainTest.php#L40

In the same test.

### DIFF
--- a/tests/Doctrine/Tests/Persistence/Mapping/DriverChainTest.php
+++ b/tests/Doctrine/Tests/Persistence/Mapping/DriverChainTest.php
@@ -24,13 +24,13 @@ class DriverChainTest extends DoctrineTestCase
                 ->method('isTransient');
 
         $driver2 = $this->createMock(MappingDriver::class);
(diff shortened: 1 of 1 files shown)


greg0ire (approved):

greg0ire:
Thanks @franmomu !

PR #172, opened 2021-04-22
Use PSR-6 compatibility classes from doctrine/cache

alcaeus:
This removes the dependency on symfony/cache and uses the PSR-6 compatibility classes included in doctrine/cache 1.11 (and 2.0 once it's released).

### DIFF
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -5,12 +5,12 @@ namespace Doctrine\Persistence\Mapping;
 use BadMethodCallException;
 use Doctrine\Common\Cache\Cache;
 use Doctrine\Common\Cache\CacheProvider;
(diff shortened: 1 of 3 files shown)


greg0ire (approved):

lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -91,7 +91,7 @@ public function setCacheDriver(?Cache $cacheDriver = null)
             throw new BadMethodCallException('Cannot convert cache to PSR-6 cache');
```
stof, line 91: should this exception be removed entirely ? AFAICT, `CacheAdapter` accepts a `Cache` as its argument (unless the code in doctrine/cache is wrong)

alcaeus: You are correct, this was left over from the previous compatibility layer. I've created #182 to fix this.

PR #171, opened 2021-04-22
Make Classmetadata generic

franmomu:
Last bit of https://github.com/doctrine/persistence/pull/159

PR #173, opened 2021-04-24
Fix annotation template and add more phpdoc

franmomu:
Reviewing again https://github.com/doctrine/persistence/pull/171, just realised that I used in test `@implements` where I meant `@template-implements`.

I raised locally the phpstan level and added some more phpdoc, but it fails:

```
InvalidArgument - tests/Doctrine/Tests/Persistence/PersistentObjectTest.php:55:49 - Argument 2 of Doctrine\Tests\Persistence\TestObject::injectObjectManager expects Doctrine\Persistence\Mapping\ClassMetadata<object>, Doctrine\Tests\Persistence\TestObjectMetadata provided (see https://psalm.dev/004)
        $this->object->injectObjectManager($om, $this->cm);
```

should `ClassMetadata` be declared as `@template-covariant`? there are some other places like:

https://github.com/doctrine/persistence/blob/fb399fefb5570c27c8b15c69c797b0ee559e1fe8/lib/Doctrine/Persistence/Event/LoadClassMetadataEventArgs.php

that I don't know how would be the best way to specify the type of `ClassMetadata` apart from `ClassMetadata<object>`, but I guess it should be declared as covariant for that to work.

### DIFF
--- a/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -313,6 +313,7 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      * @psalm-param class-string $name
      *
      * @return string[]
+     * @psalm-return class-string[]
      */
     protected function getParentClasses($name)
     {
@@ -428,9 +429,12 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      * Creates a new ClassMetadata instance for the given class name.
      *
      * @param string $className
+     * @psalm-param class-string<T> $className
      *
-     * @return ClassMetadata
+     * @return ClassMetadata<T>
      * @psalm-return CMTemplate
+     *
+     * @template T of object
      */
     abstract protected function newClassMetadataInstance($className);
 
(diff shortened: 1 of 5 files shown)


greg0ire:
@orklah can you please advise?

orklah:
I'm not going to be much helpful here. I still have a hard time wrapping my head around that

greg0ire:
Ok, no worries :)

I just read up on `template-covariant` here, and from what I understand, it should be used when it is fine to use subtypes instead of objects. I don't think we can guarantee that, but we can be sure that only subtypes of `object` are ever going to be used. So I think we don't really have a choice here. Let's use it!

greg0ire:
@franmomu a release is coming up, let's try with the `template-covariant` annotation?

franmomu:
> @franmomu a release is coming up, let's try with the `template-covariant` annotation?

Sure! I was waiting if someone else wanted to chime in, I'll do it this morning.

franmomu:
@greg0ire a merge up is needed from `2.1.x`, I think

greg0ire:
I'm AFK for 1 week

franmomu:
> I'm AFK for 1 week

Enjoy! 👍

alcaeus:
Moving this to 2.3 so I can get on with the cache changes.

greg0ire:
Restarting the build since a merge up happened in the meantime

greg0ire:
Looks like more work is needed

UPD: actually, the same errors can be seen at https://github.com/doctrine/persistence/pull/177

greg0ire (approved):

greg0ire:
Thanks @franmomu !

PR #180, opened 2021-05-16
Phpstan lvl 7

greg0ire:


greg0ire:
Let's wait for https://www.githubstatus.com/incidents/zbpwygxwb3gw to be resolved before merging this.

CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
+# Circular dependency
+
+This package has a development dependency on `doctrine/common`, which has a
```
alcaeus, line 3: Thanks for documenting this. Is there anything we can do to fix this? Should we prepare doctrine/common to work with persistence 3?

alcaeus (approved):
Can't comment on the impact of the psalm changes, but in general LGTM.

CONTRIBUTING.md
greg0ire: We could use `^2.0 || ^3.0`, but IMO that's a bit "risky" if we do not test that 2.0 actually work with 3, and the benefit is small: that would only fix one weird thing out of the 2 we have to do, namely:
- using `COMPOSER_ROOT_VERSION`
- using `2.0` as a value for it when developing on 3.x

PR #183, opened 2021-05-17
Describe some types more accurately

greg0ire:
[Downstream builds are broken](https://github.com/doctrine/orm/pull/8546#issuecomment-842465855) partly because the last release of persistence
describes its inputs more accurately, but not its outputs.


lib/Doctrine/Persistence/Mapping/Driver/DefaultFileLocator.php
```diff
@@ -136,7 +136,9 @@ public function getAllClassNames($globalBasename)
                     }
 
                     // NOTE: All files found here means classes are not transient!
-                    $classes[] = str_replace('.', '\\', $fileName);
+                    /** @psalm-var class-string */
```
greg0ire, line 139: Using `assert(class_exists($class))` on this particular one makes the ORM test suite fail https://github.com/doctrine/orm/blob/2.8.x/tests/Doctrine/Tests/ORM/Tools/ConvertDoctrine1SchemaTest.php because there is heavy mocking involved.

greg0ire:
@orklah @franmomu please review. More things could be done if it were not for https://github.com/vimeo/psalm/issues/5788

orklah:
I think a good amount of the arrays could be described as `list<class-string>` if we wanted to. Otherwise, seems good!

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -61,6 +61,7 @@ abstract class AnnotationDriver implements MappingDriver
      * Cache for AnnotationDriver#getAllClassNames().
      *
      * @var string[]|null
+     * @psalm-var list<class-string>|null
```
alcaeus, line 64: Nit: if this is an array, we could initialise it to `[]` and avoid `null`, but I'll admit I haven't checked if the code has any special handling when it is indeed `null`. We can also make the change in 2.3 if you don't want to make this kind of change in a patch release.

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
```diff
@@ -132,10 +134,10 @@ public function getAllClassNames()
             return (array) $this->locator->getAllClassNames($this->globalBasename);
         }
 
-        return array_unique(array_merge(
+        return array_values(array_unique(array_merge(
```
alcaeus, line 137: What's the reason for adding `array_values` here? I thought `array_unique` renumbers arrays?

lib/Doctrine/Persistence/Mapping/Driver/MappingDriver.php
```diff
@@ -13,6 +13,7 @@ interface MappingDriver
      * Loads the metadata for the specified class into the provided container.
      *
      * @param string $className
+     * @psalm-param class-string $className
```
franmomu: I guess here `@template T of object` could be used to specify `class-string<T>` and `ClassMetadata<T>`.

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
greg0ire: Apparently it doesn't:

```
>>> array_unique([2 => 3])
=> [
     2 => 3,
   ]
```

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
greg0ire: There is in fact special handling here: 

https://github.com/doctrine/persistence/blob/8b420f7065c2d6c0149ff523816f6b28391581d1/lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php#L194-L196

I must admit I didn't check the code either before you asked.

lib/Doctrine/Persistence/Mapping/Driver/MappingDriver.php
greg0ire: Oh right, fixed, thanks!

greg0ire: I don't really understand how I can fix the resulting error, maybe I should ignore it…

lib/Doctrine/Persistence/Mapping/Driver/FileDriver.php
alcaeus: Wow. I see.

lib/Doctrine/Persistence/Mapping/Driver/MappingDriver.php
franmomu: hmmm I guess so, what I don't know is why it didn't complain calling `object::loadMetadata` 🤔

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
alcaeus: Then let's not make the change here but in a separate PR 👍

lib/Doctrine/Persistence/Mapping/Driver/MappingDriver.php
greg0ire: https://github.com/phpstan/phpstan/issues/5023

alcaeus (approved):

alcaeus:
Thanks @greg0ire!

PR #192, opened 2021-06-15
Allow null as a return value for association's target

malarzm:
This came up in https://github.com/doctrine/mongodb-odm/issues/2325 (cc @franmomu @alcaeus ). For the background:

1. ODM has a concept of a discriminator map for references which allows to specify multiple classes a reference can point to, this setting must not be combined with `targetDocument`
2. In ODM you can omit `targetDocument` and `discriminatorMap` altogether to have a reference to any mapped document. Behind the scenes ODM is using FQCN as a discriminator's value 

The change as-is is needed to prevent a break when doctrine/persistence becomes typed. The other solution I see is allowing the method to return an array of FQCNs that would allow ODM to reflect discriminator maps in the method's return value. This would cause a minor inconvenience for ORM I believe, but nothing unseen before (I'm looking at `isSingleValuedAssociation` and similar thing for identifiers).

### DIFF
--- a/lib/Doctrine/Common/Persistence/PersistentObject.php
+++ b/lib/Doctrine/Common/Persistence/PersistentObject.php
@@ -109,7 +109,7 @@ abstract class PersistentObject implements ObjectManagerAware
             $this->$field = $args[0];
         } elseif ($this->cm->hasAssociation($field) && $this->cm->isSingleValuedAssociation($field)) {
             $targetClass = $this->cm->getAssociationTargetClass($field);
-            if (! ($args[0] instanceof $targetClass) && $args[0] !== null) {
+            if ($targetClass !== null && ! ($args[0] instanceof $targetClass) && $args[0] !== null) {
                 throw new InvalidArgumentException("Expected persistent object of type '" . $targetClass . "'");
             }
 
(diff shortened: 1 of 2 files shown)


alcaeus:
No objection to the change in general, but I feel like the two PHPStan errors point to potential downstream breaks. @greg0ire @franmomu what do you think?

greg0ire:
I think it looks a bit like: https://github.com/doctrine/orm/pull/8767#discussion_r651630101

Now that we are able to add type information through phpdoc and check it with SA tools, there is going to be a phase where we notice all the inconsistencies in how we implement our interfaces.

It is still compatible to say the implementation in the ORM still returns something non-nullable, so here the solution to the breakage will be to override the signature to just say that.

franmomu:
> It is still compatible to say the implementation in the ORM still returns something non-nullable, so here the solution to the breakage will be to override the signature to just say that.

I agree with this, I guess overriding it in the ORM would be enough, I don't think many packages rely on `ClassMetadata` of persistence.

By the way, apparently `PersistenObject` should have been removed long time ago.

alcaeus (commented):
LGTM with static analysis failures addressed: we need to explicitly check for `null` before using `instanceof`

greg0ire (approved):

alcaeus:
Thanks @malarzm!

VincentLanglet:
@greg0ire I now have a phpstan error, using orm, because `class-string|null` can be returned.

Does https://github.com/doctrine/orm/blob/2.9.x/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L3414 can really return `null` ? If not, phpdoc should be added.

stof:
Based on the mapping validation in the same class, it cannot be `null`

VincentLanglet:
I opened https://github.com/doctrine/orm/pull/8907 then

PR #197, opened 2021-06-25
Mark doctrine/annotations dependency as optional

### LINKED ISSUE #195: Don't define Doctrine annotations as a hard dependency?
Similar to https://github.com/doctrine/orm/issues/8785, this project defines `doctrine/annotations` as a required dependency:

https://github.com/doctrine/persistence/blob/ee36ff348ea134f5d2ca4745368307e4629109bb/composer.json#L24

But since https://github.com/doctrine/orm/pull/8266 it's possible to define entity mapping with PHP attributes instead of annotations. Thanks.

greg0ire:
Now that the ORM has a driver based on PHP attributes, it seems less
necessary to require doctrine/annotations. In fact, that could also be
considered true before since there are already other drivers than the
annotation driver.
Fixes #195

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,6 @@
     ],
     "require": {
         "php": "^7.1 || ^8.0",
-        "doctrine/annotations": "^1.0",
         "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/collections": "^1.0",
         "doctrine/event-manager": "^1.0",
@@ -31,6 +30,7 @@
     "require-dev": {
         "composer/package-versions-deprecated": "^1.11",
         "phpstan/phpstan": "0.12.84",
+        "doctrine/annotations": "^1.0",
         "doctrine/coding-standard": "^6.0 || ^9.0",
         "doctrine/common": "^3.0",
         "phpunit/phpunit": "^7.5.20 || ^8.0 || ^9.0",
@@ -38,6 +38,7 @@
         "vimeo/psalm": "4.7.0"
     },
     "conflict": {
+        "doctrine/annotations": "<1.0 || >=2.0",
         "doctrine/common": "<2.10@dev"
     },
     "autoload": {


greg0ire:
Cc @javiereguiluz :slightly_smiling_face:

alcaeus (dismissed):

composer.json
```diff
@@ -31,13 +30,15 @@
...
         "phpstan/phpstan": "0.12.84",
+        "doctrine/annotations": "^1.0",
         "doctrine/coding-standard": "^6.0 || ^9.0",
         "doctrine/common": "^3.0",
         "phpunit/phpunit": "^7.5.20 || ^8.0 || ^9.0",
         "symfony/cache": "^4.4|^5.0",
         "vimeo/psalm": "4.7.0"
     },
     "conflict": {
+        "doctrine/annotations": "<1.0 || ^2.0",
```
stof: if the goal is to allow on 1.x, the second part of the conflict should be `>=2.0`, not `^2.0`. the current conflict rule allows 3.0.

alexislefebvre:
If we upgrade from 2.2.4 to 2.3.0, `doctrine/annotations` is not installed anymore. So that `Doctrine\Common\Annotations\AnnotationReader` is not available anymore. Is it a BC break or a weak dependency in the projects that relied on `AnnotationReader` through `doctrine/persistence`?

Here is how I fixed that *issue* in one project: https://github.com/liip/LiipTestFixturesBundle/pull/170/files

derrabus:
If your project relies on Doctrine Annotations for metadata mapping, adding the explicit dependency is the way to go.

derrabus:
Related: doctrine/orm#9452

alexislefebvre:
Thanks, so I wonder how `AnnotationReader` can be available [here in FrameworkBundle](https://github.com/symfony/symfony/blob/a061fc4a3db54ceaf3b3fd821d35beb047d967f1/src/Symfony/Bundle/FrameworkBundle/Resources/config/annotations.php#L14) when `doctrine/annotations` is only a [`dev` dependency](https://github.com/symfony/symfony/blob/a061fc4a3db54ceaf3b3fd821d35beb047d967f1/src/Symfony/Bundle/FrameworkBundle/composer.json#L36)

Update: my bad, it's a `dev` dependency since Symfony 4.4 at least, so this is not related to this PR.

derrabus:
A `use` statement alone does not require the referenced class to exist.

alexislefebvre:
I'll have to investigate on Symfony's side, but I ended up with a `annotations.reader` service that was missing because `Doctrine\Common\Annotations\AnnotationReader` was not installed.

derrabus:
Yes. If the library is missing, Symfony won't create that service. But let's not flood this PR with chat messages. Feel free to open a discussion on either board, if you have further questions:

* https://github.com/symfony/symfony/discussions
* https://github.com/doctrine/orm/discussions

PR #198, opened 2021-06-26
Remove hard dependency on doctrine/annotations

wouterj:
It would be great to be able to remove `doctrine/annotations` in projects that fully migrated to attributes. While testing this, we discovered that this package has a hard dependency on it (and thus, it cannot be removed).

The diff of this PR feels a bit weird: I couldn't find any dependency on `doctrine/annotations` in this project (other than the tests).

The `AnnotationDriver` does have a PHPdoc-only typehint to `Doctrine\Common\Annotations\Reader`. The `doctrine/orm` `AttributeDriver` extends this class, but passes a custom `AttributeReader` that doesn't extend from the annotations reader. So I guess that's a reason why the typehint isn't a PHP typehint?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,6 @@
     ],
     "require": {
         "php": "^7.1 || ^8.0",
-        "doctrine/annotations": "^1.0",
         "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/collections": "^1.0",
         "doctrine/event-manager": "^1.0",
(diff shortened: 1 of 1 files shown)


javiereguiluz:
Thanks Wouter! ... but this change was already merged in #197.

wouterj:
Oh, sorry! Thanks for letting me know!

PR #202, opened 2021-08-10
Add explicit `@return` type next to `#[ReturnTypeWillChange]`

nicolas-grekas:
Same as https://github.com/doctrine/collections/pull/280


malarzm (approved):

malarzm:
Thanks @nicolas-grekas!

PR #208, opened 2021-10-23
Undeprecate detach

malarzm:
Let's finally get `detach` off the hook. As discussed we can not deprecate detach without providing a viable alternative and using `clear` isn't one of them.


stof (approved):

greg0ire:
Regarding the failing build: https://github.com/doctrine/persistence/pull/209

greg0ire (approved):

malarzm:
I'll let it be open until Monday and then merge given no objections are raised

PR #229, opened 2022-02-05
Document 3.0.0 BC breaks

greg0ire:
Note that some BC Breaks seem to have wrongly been documented in
UPGRADE-2.0.md


derrabus (approved):

derrabus (changes requested):
I think the whole UPGRADE-2.0.md should go away. It does not exist on 2.3.x.

greg0ire:
Good catch lol… right when I was merging.

derrabus (approved):

PR #230, opened 2022-02-06
Revert return type declarations

greg0ire:
Adding them means downstream projects that define subtypes of ours
cannot create subtypes compatible with both APIs without a BC break, as
they would need to add return types as well, which is a BC break unless
the subtype at hand is a final class.

To make sure I did not do less or more than I should, I used `git diff origin/2.4.x..revert-return-types -S '(public|protected).*\)\:' --pickaxe-regex -p lib/`

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -20,7 +20,9 @@ Removed support for using doctrine/cache for metadata caching. The
 + public function clear(): void
 ```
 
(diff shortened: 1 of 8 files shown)


derrabus (dismissed):

derrabus:
Rebase needed.

derrabus (approved):

PR #234, opened 2022-02-16
Improve documentation of ManagerRegistry::getManagerNames()

### LINKED ISSUE #233: Unexpected response returned from `AbstractManagerRegistry::getManagerNames()`
The `ManagerRegistry::getManagerNames()` should return `An array of object manager names. string[]`

https://github.com/doctrine/persistence/blob/0130920efd696cf5480744f499672b2d4ca47c85/lib/Doctrine/Persistence/ManagerRegistry.php#L65-L70

But the implementation in `AbstractManagerRegistry` returns an array of serviceId indexed by name.

https://github.com/doctrine/persistence/blob/0130920efd696cf5480744f499672b2d4ca47c85/lib/Doctrine/Persistence/AbstractManagerRegistry.php#L194-L197

I'm not sure what is the expected response. I would expect that `getManagerNames` returned `['foo', 'bar']`. I would expect such code worked: But this is not the case.

```php
foreach ($this->registry->getManagerNames() as $name) {
  dump($this->registry->getManager($name));
}
```

But fixing it in the `AbstractManagerRegistry` class could be a BC break.
Otherwise, we should improve the documentation of the interface.

greg0ire:
Closes #233

### DIFF
--- a/lib/Doctrine/Persistence/ManagerRegistry.php
+++ b/lib/Doctrine/Persistence/ManagerRegistry.php
@@ -63,9 +63,12 @@ interface ManagerRegistry extends ConnectionRegistry
     public function getAliasNamespace($alias);
 
     /**
-     * Gets all object manager names.
+     * Gets all object manager names and associated service IDs. A service ID
+     * is a string that allows to obtain an object manager, typically from a
+     * PSR-11 container.
      *
-     * @return string[] An array of object manager names.
+     * @return array<string,string> An array with object manager names as keys,
+     *                              and service IDs as values.
      */
     public function getManagerNames();
 


lib/Doctrine/Persistence/ManagerRegistry.php
```diff
@@ -65,7 +65,8 @@ public function getAliasNamespace($alias);
     /**
      * Gets all object manager names.
      *
-     * @return string[] An array of object manager names.
+     * @return array<string,string> An array with object manager names as keys,
+     *                              and service ids as values.
```
derrabus: What's a "service id"? It's the first time the concept of a service appears on that contract.

greg0ire: I hesitated to mention a `PSR-11` service but technically I don't think it has to be. The only other description of this concept in the package is here: https://github.com/doctrine/persistence/blob/0130920efd696cf5480744f499672b2d4ca47c85/lib/Doctrine/Persistence/AbstractManagerRegistry.php#L58-L67

Not sure if I should be more precise in the interface.

derrabus: I think we should add a short paragraph to the doc block of this method, about what a service is in our context. Right now, the sudden mention of service IDs is a bit confusing.

derrabus (approved):

PR #239, opened 2022-02-21
Assert AnnotationDriver works with winding paths

greg0ire:
This has been broken on the 3.0.x branch. Let us add a test that ensure
this cannot be broken.

### DIFF
--- a/lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
@@ -220,7 +220,7 @@ abstract class AnnotationDriver implements MappingDriver
             foreach ($iterator as $file) {
                 $sourceFile = $file[0];
 
-                if (! preg_match('(^phar:)i', $sourceFile)) {
+                if (preg_match('(^phar:)i', $sourceFile) === 0) {
                     $sourceFile = realpath($sourceFile);
                 }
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -220,7 +220,7 @@ public function getAllClassNames()
             foreach ($iterator as $file) {
                 $sourceFile = $file[0];
 
-                if (! preg_match('(^phar:)i', $sourceFile)) {
+                if (preg_match('(^phar:)i', $sourceFile) === 0) {
```
greg0ire, line 223: #50 wrongly uses `=== false` here. `false` is returned if an error occurs, the correct to way to assert that there is no match is to count them and assert that that count is 0.

tests/Doctrine/Tests/Persistence/Mapping/AnnotationDriverTest.php
```diff
@@ -20,6 +24,15 @@ public function testGetAllClassNames(): void
...
         self::assertSame([TestClass::class], $classes);
     }
+
+    /**
+     * @return Generator<string, array{string}>
+     */
+    public function pathProvider(): Generator
+    {
+        yield 'straigthforward path' => [__DIR__ . '/_files/annotation'];
+        yield 'winding path' => [__DIR__ . '/../_files/annotation'];
```
derrabus:
```suggestion
        yield 'winding path' => [__DIR__ . '/../Mapping/_files/annotation'];
```

greg0ire: Oh right, I meant to type this actually :sweat_smile:

derrabus (approved):

derrabus:
The fun of merging this up to 3.0 is all yours this time. 🙈

PR #246, opened 2022-02-26
Decouple from annotations

greg0ire:
Closes #217


lib/Doctrine/Persistence/Mapping/Driver/StaticPHPDriver.php
```diff
@@ -68,7 +68,8 @@ public function loadMetadataForClass($className, ClassMetadata $metadata)
     /**
      * {@inheritDoc}
      *
-     * @todo Same code exists in AnnotationDriver, should we re-use it somehow or not worry about it?
+     * @todo Same code exists in ColocatedMappingDriver, should we re-use it
+     * somehow or not worry about it?
```
greg0ire, line 72: This is an interesting comment. The code seems to have diverged a bit though.

lib/Doctrine/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -0,0 +1,209 @@
...
+use function preg_match;
+use function preg_quote;
+use function realpath;
+use function str_replace;
+use function strpos;
+
+/**
+ * The ColocatedMappingDriver reads the mapping metadata located near the code.
+ */
+abstract class ColocatedMappingDriver implements MappingDriver
```
derrabus: I wonder if we should make this a trait instead.

`ColocatedMappingDriver` as a type does not make sense; We'll probably never see `ColocatedMappingDriver` in a return type or `instanceof` check for instance. And all this class contains is just some shared implementation.

greg0ire: Good point, I hadn't considered that. The fact that that class is hard to name is probably a hint at the fact that there should not be a type.

greg0ire:
ping @derrabus :pray:

lib/Doctrine/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -0,0 +1,215 @@
...
+     */
+    protected $classNames;
+
+    /**
+     * @param string ...$paths One or multiple paths where mapping classes can be found.
+     */
+    public function __construct(string ...$paths)
+    {
+        $this->addPaths($paths);
+    }
```
derrabus: I would not move the constructor into the trait. `AnnotationDriver` for instance skips it already.

lib/Doctrine/Persistence/Mapping/MappingException.php
```diff
@@ -28,12 +28,16 @@ public static function classNotFoundInNamespaces($className, $namespaces)
     }
 
     /**
+     * @param class-string $driverClassName
+     *
      * @return self
      */
-    public static function pathRequired()
+    public static function pathRequired(string $driverClassName)
```
derrabus: This is a BC break. Shall we add a new named constructor (e.g. `pathRequiredForDriver`) instead?

tests/Doctrine/Tests/Persistence/Mapping/ColocatedMappingDriverTest.php
```diff
@@ -0,0 +1,54 @@
...
+use PHPUnit\Framework\TestCase;
+
+class ColocatedMappingDriverTest extends TestCase
+{
+    /**
+     * @dataProvider pathProvider
+     */
+    public function testGetAllClassNames(string $path): void
+    {
+        $driver = new MyDriver($path);
```
derrabus: Shall we move the constructor call into a protected method that `AnnotationDriverTest` can override?

lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -82,57 +40,10 @@ abstract class AnnotationDriver implements MappingDriver
     public function __construct($reader, $paths = null)
```
derrabus, line 40: Should the constructor trigger a runtime deprecation?

derrabus (changes requested):

derrabus (approved):

docs/en/reference/index.rst
```diff
@@ -174,8 +174,8 @@ That may be in files, annotations, yaml, xml, etc.
         public function isTransient($className);
     }
 
-The Doctrine Persistence project offers a few base implementations that make it easy to implement your own XML,
-Annotations or YAML drivers.
+The Doctrine Persistence project offers a few base implementations that
+make it easy to implement your own XML, Attributes or YAML drivers.
```
malarzm, line 178: Attributes, or YAML

docs/en/reference/index.rst
```diff
@@ -233,45 +233,9 @@ AnnotationDriver
 
 .. note::
 
-    This driver requires the ``doctrine/annotations`` project. You can install it with composer.
+    This driver requires the ``doctrine/annotations`` project and is
+    deprecated because of that.
```
malarzm, line 237: I'd add a note about attributes superseding annotations

malarzm: Also why removing the docs below? The feature is deprecated but documentation is still valuable :)

lib/Doctrine/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -0,0 +1,207 @@
...
+use function preg_match;
+use function preg_quote;
+use function realpath;
+use function str_replace;
+use function strpos;
+
+/**
+ * The ColocatedMappingDriver reads the mapping metadata located near the code.
+ */
+trait ColocatedMappingDriver
```
malarzm, line 28: why trait instead of an abstract class?

malarzm (commented):
I'm sorry I'm late to the party, left few minor comments

lib/Doctrine/Persistence/Mapping/Driver/ColocatedMappingDriver.php
derrabus: There's a resolved comment about that in this PR.

malarzm: Ah thanks, haven't seen that :)

docs/en/reference/index.rst
greg0ire: I think it feels weird to deprecate a feature yet help people build on it. I thought I would make it harder for people to shoot themselves in the foot.

PR #247, opened 2022-02-26
Use a shallower tree

greg0ire:
It makes it easier to work with fuzzy matchers.
I'm targeting the lowest supported branch because otherwise some operations might be painful, like switching branches on Github while browsing a file, or performing a merge up in some situation.

I'll take care of merging this up.


SenseException (dismissed):
Technically okay, but where is fuzzy matching used that it affects the changes?

greg0ire:
> Technically okay, but where is fuzzy matching used that it affects the changes?

You mean that it is affected by the changes? [Fuzzy matching is used in IDEs](https://github.com/nvim-telescope/telescope.nvim) and works best in my experience when file paths are short and not repetitive.

derrabus:
I'd feel more comfortable merging this to 2.4.x, tbh. This is certainly not a bugfix.

greg0ire:
@derrabus OK, I'll retarget. It might be cumbersome for a while, but 2.3.x isn't going to live very long.

derrabus:
Let's finish the other work planned for 2.4 first an then rebase an merge this PR.

malarzm (dismissed):

greg0ire:
I think 2.4.0 should be released, so let's revive this?

SenseException (approved):

PR #248, opened 2022-02-28
add enum reflection property

IonBazan:
Reduces code duplication in https://github.com/doctrine/mongodb-odm/pull/2412 and https://github.com/doctrine/orm/pull/9304 by extracting the `EnumReflectionProperty` here. Please take note of slightly changed name and throwing native `ValueException` here instead of `MappingException`.

### DIFF
--- a/tests/Doctrine/Tests_PHP81/Persistence/Reflection/EnumReflectionPropertyTest.php
+++ b/tests/Doctrine/Tests_PHP81/Persistence/Reflection/EnumReflectionPropertyTest.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Doctrine\Tests_PHP81\Persistence\Reflection;
+
+use Doctrine\Persistence\Reflection\EnumReflectionProperty;
+use PHPUnit\Framework\TestCase;
+use ReflectionProperty;
+use ValueError;
+
+class EnumReflectionPropertyTest extends TestCase
+{
+    public function testGetValue(): void
+    {
+        $object       = new TypedEnumClass();
+        $reflProperty = new EnumReflectionProperty(new ReflectionProperty(TypedEnumClass::class, 'suit'), Suit::class);
+        self::assertNull($reflProperty->getValue($object));
+        $object->suit = Suit::Clubs;
+        self::assertSame('C', $reflProperty->getValue($object));
+        $object->suit = null;
+        self::assertNull($reflProperty->getValue($object));
+    }
+
+    public function testSetValidValue(): void
+    {
+        $object       = new TypedEnumClass();
+        $object->suit = Suit::Hearts;
+
+        $reflProperty = new EnumReflectionProperty(new ReflectionProperty(TypedEnumClass::class, 'suit'), Suit::class);
+
+        $reflProperty->setValue($object);
+        self::assertNull($reflProperty->getValue($object));
+        /** @psalm-suppress TypeDoesNotContainNull */
+        self::assertNull($object->suit);
+
+        $reflProperty->setValue($object, 'D');
+        self::assertSame('D', $reflProperty->getValue($object));
+        self::assertSame(Suit::Diamonds, $object->suit);
+    }
+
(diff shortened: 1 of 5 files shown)


tests/Doctrine/Tests_PHP81/Persistence/Reflection/EnumReflectionPropertyTest.php
```diff
@@ -0,0 +1,67 @@
...
+    public ?Suit $suit = null;
+}
+
+enum Suit: string
+{
+case Hearts = 'H';
+case Diamonds = 'D';
+case Clubs = 'C';
+case Spades = 'S';
+    }
```
derrabus:
```suggestion
enum Suit: string
{
    case Hearts = 'H';
    case Diamonds = 'D';
    case Clubs = 'C';
    case Spades = 'S';
}
```

To convince PHPCS to accept this indentation, you have to disable certain rules for this file.

```XML
<rule ref="Generic.WhiteSpace.ScopeIndent.Incorrect">
    <!-- see https://github.com/squizlabs/PHP_CodeSniffer/issues/3474 -->
    <exclude-pattern>path/to/your/file</exclude-pattern>
</rule>
<rule ref="Generic.WhiteSpace.ScopeIndent.IncorrectExact">
    <!-- see https://github.com/squizlabs/PHP_CodeSniffer/issues/3474 -->
    <exclude-pattern>path/to/your/file</exclude-pattern>
</rule>
```

IonBazan: Applied, thanks!

tests/Doctrine/Tests_PHP81/Persistence/Reflection/EnumReflectionPropertyTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+        self::assertNull($reflProperty->getValue($object));
+
+        $object->suit = Suit::Clubs;
+
+        self::assertSame('C', $reflProperty->getValue($object));
+
+        $object->suit = null;
+
+        self::assertNull($reflProperty->getValue($object));
```
greg0ire, line 20: Skipping lines is usually used to group lines together in a block. Here, nothing is grouped, so why even have blank lines?

greg0ire:
> throwing native ValueException here instead of MappingException.

I think you mean `ValueError`. Besides, since `doctrine/persistence` has its own `MappingException` class, why not preserve the original behavior and wrap the `ValueError` in it?

malarzm:
> I think you mean ValueError. Besides, since doctrine/persistence has its own MappingException class, why not preserve the original behavior and wrap the ValueError in it?

I'd say `MappingException` is for mapping errors :) I'm perfectly fine letting `ValueError` be thrown, for instance in ODM we'll catch it and wrap in our `HydratorException`: https://github.com/doctrine/mongodb-odm/pull/2412#discussion_r821882889

malarzm (approved):

greg0ire:
@malarzm I wasn't worried so much about preserving the type, which wouldn't even be achieved by my proposal, as about preserving the extra context that is added:

```
Context: Trying to hydrate enum property "%s::$%s"
Problem: Case "%s" is not listed in enum "%s"
Solution: Either add the case to the enum type or migrate the database column to use another case of the enum
```

If you have many entities/document using the same enum, and your database/document store contains an invalid value for that enum somewhere, it might help to know where.

`HydratorException` does seem like a better type for this, maybe we should introduce something similar in `doctrine/persistence`?

malarzm:
We'll totally do what you're proposing, but on ODM's level. I'm not sure how much value the exception will bring in this case. For instance in ODM the `HydratorException` is already extending our base exception and we can't change that. So we'd need to catch the new exception and later transform it into our one. Not sure if it's worth it :)

greg0ire:
Thanks @IonBazan !

PR #249, opened 2022-03-13
Merge 2.4.x up into 3.0.x

greg0ire:


UPGRADE.md
```diff
@@ -0,0 +1,90 @@
...
+- Use of `@deprecated` docblock that is detected by IDEs (like PHPStorm) or
+  Static Analysis tools (like Psalm, phpstan)
+- Use of our low-overhead runtime deprecation API, details:
+  https://github.com/doctrine/deprecations/
+
+# Upgrade to 3.0
+
+## BC Break: removed `ObjectManager::merge()`
+
+`ObjectManagerDecorator::merge()` is removed as well.
```
malarzm, line 13:
```suggestion
`ObjectManagerDecorator::merge()` is removed without replacement.
```

I believe "as well" part originates from times we wanted to remove detach, but since we're not let's rephrase

UPGRADE.md
```diff
@@ -0,0 +1,90 @@
...
+
+Please implement this functionality directly in your application if you want
+ActiveRecord style functionality.
+
+# Upgrade to 2.4
+
+## Deprecated `AnnotationDriver`
+
+Since attributes were introduced in PHP 8.0, annotations are deprecated. Use
+`ColocatedMappingDriver` instead. This will involve implementing
```
malarzm, line 45: I should have asked in the PR but here I am now - is it right to advertise a trait here without further explaination?

src/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -0,0 +1,74 @@
...
+
+use Doctrine\Common\Annotations\Reader;
+use ReflectionClass;
+
+use function get_class;
+
+/**
+ * The AnnotationDriver reads the mapping metadata from docblock annotations.
+ *
+ * @deprecated extend ColocatedMappingDriver directly instead.
```
malarzm, line 15: This also talks about extension, not using the trait.

malarzm (approved):

PR #254, opened 2022-03-19
Drop deprecated APIs

greg0ire:


UPGRADE.md
```diff
@@ -8,6 +8,20 @@ awareness about deprecated code.
 
 # Upgrade to 3.0
 
+## BC Break: removed `AnnotationDriver`
+
+It has been deprecated in favor of `ColocatedMappingDriver`.
```
malarzm: s/deprecated/removed everywhere I think :)

tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -49,15 +50,15 @@ public function testSetGetCache(): void
...
 
-        self::assertTrue($this->cmf->hasMetadataFor('stdClass'));
+        self::assertTrue($this->cmf->hasMetadataFor(stdClass::class));
     }
 
     public function testGetMetadataForAbsentClass(): void
     {
         $this->expectException(MappingException::class);
-        $this->cmf->getMetadataFor(__NAMESPACE__ . '\AbsentClass');
+        $this->cmf->getMetadataFor(Foo::class);
```
malarzm, line 61: Staying with the former wouldn't produce a new PHPStan's baseline entry

malarzm (dismissed):

tests/Persistence/Mapping/ClassMetadataFactoryTest.php
greg0ire: In isolation, yes, but there are other occurrences of `Foo::class` in that file.

UPGRADE.md
greg0ire: What I mean is: we removed `AnnotationDriver` because it has been deprecated in favor of `ColocatedMappingDriver`.

malarzm: Ah, I've read that the other way. Anyway it's inconsistent with next paragraphs this way :)

greg0ire: I used another phrasing 👍

malarzm (approved):

PR #259, opened 2022-03-22
Missed tests exports ignore

a-menshchikov:


greg0ire (approved):

greg0ire:
Thanks @a-menshchikov !

a-menshchikov:
@greg0ire could you tell when this code might be released?
Without it Psalm throws a ParseError for PHP 8.0 (or older) projects. It can be suppressed, but much better to exclude a root cause.

greg0ire:
:thinking: why does Psalm need to parse test files?

UPD: https://github.com/doctrine/persistence/releases/tag/2.4.1

a-menshchikov:
@greg0ire 
Thanks for release! :+1: 

To be honest, I can't understand it myself. I catch this error even if I add the whole "vendor" directory in a "projectFiles > ignoreFiles". At the same time I have "vendor" in an "extraFiles" property, maybe it depends.

greg0ire:
Loading and inspecting seem to be 2 different things, and `ignoreFiles` probably only applies to the latter. The description of `extraFiles` is the following:

> Directories Psalm should load but not inspect.

I can't find something to prevent some files being loaded

PR #263, opened 2022-03-27
Stop using ObjectManagerAware in tests

greg0ire:
It also seems some tests were invalid: ObjectManagerAware was used to
avoid creating a custom Proxy interface but really is not. TestObject is
not a proxy, and should not implement it.

To understand this, I found a commit but no pull request in this repository, then I remembered this repository was split from another, so here it is: https://github.com/doctrine/common/pull/387

### DIFF
--- a/tests/Persistence/ManagerRegistryTest.php
+++ b/tests/Persistence/ManagerRegistryTest.php
@@ -8,7 +8,6 @@ use Doctrine\Persistence\AbstractManagerRegistry;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Persistence\ObjectManager;
-use Doctrine\Persistence\ObjectManagerAware;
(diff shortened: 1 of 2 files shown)


greg0ire:
I'm not sure if this is the correct thing to do, so I'd like to summon some help in untangling this:

- @Ocramius and @enumag because of the `doctrine/common` PR
- @lcobucci because of 7538983699183d79dc2854953609359b2c934008, which introduced the calls to `assertNull()`
- @phansys because of f045d3cf58b48008ef997a9f287641ecf9c52c86, which introduced a test without assertions.

lcobucci:
@greg0ire I vaguely remember debugging this... I think the assertions were added because that was the return back then (it's been a while, though)

greg0ire:
Ok, then it's probably fine to remove them.

derrabus (approved):

lcobucci:
I'd use different a approach for checking that (`addToAssertionCount` or something), though. That annotation makes phpunit ignore the test coverage - which influences mutation tests.

PR #270, opened 2022-03-29
Deprecate methods related to partial clearing

greg0ire:
Partially clearing the object manager has been deprecated, which means
these methods will become useless in 3.0

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,13 @@ awareness about deprecated code.
 
 # Upgrade to 2.5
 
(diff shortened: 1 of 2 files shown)


malarzm (dismissed):

derrabus:
Passing `$entityClass !== null` to the constructor should also be deprecated.

greg0ire:
@derrabus I also deprecated passing `null`, since it would not be compatible with the signature of 3.0

derrabus (approved):

PR #279, opened 2022-04-13
Split MappingException

greg0ire:
This will allow us to make the MappingException class an interface in
3.0. Additionally, it allows client code to catch exceptions more
precisely, and makes room for extra named constructors.

Prompted by https://github.com/doctrine/orm/pull/7199#issuecomment-1097249126

### DIFF
--- a/src/Persistence/Mapping/Exception/ClassNotFoundInNamespaces.php
+++ b/src/Persistence/Mapping/Exception/ClassNotFoundInNamespaces.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Persistence\Mapping\Exception;
+
+use function implode;
+use function sprintf;
+
+final class ClassNotFoundInNamespaces extends MappingException
+{
+    /**
+     * @param class-string $className
+     * @param string[]     $namespaces
+     */
+    public static function create(string $className, array $namespaces): self
+    {
+        return new self(sprintf(
+            'The class "%s" was not found in the chain of configured namespaces %s',
+            $className,
+            implode(', ', $namespaces)
+        ));
+    }
+}
--- a/src/Persistence/Mapping/Exception/MappingException.php
+++ b/src/Persistence/Mapping/Exception/MappingException.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Doctrine\Persistence\Mapping\Exception;
+
+use Doctrine\Persistence\Mapping\MappingException as LegacyMappingException;
+
+/**
+ * Will become an interface in 3.0.
+ * Do not extend or instantiate this class outside this package.
+ */
+class MappingException extends LegacyMappingException
+{
+}
--- a/tests/Persistence/Mapping/ClassMetadataFactoryTest.php
+++ b/tests/Persistence/Mapping/ClassMetadataFactoryTest.php
@@ -136,7 +136,7 @@ class ClassMetadataFactoryTest extends DoctrineTestCase
 
         $this->expectException(MappingException::class);
         $this->expectExceptionMessage(
-            'Class \'Doctrine\Tests\Persistence\Mapping\ChildEntity:Foo\' does not exist'
+            'Class "Doctrine\Tests\Persistence\Mapping\ChildEntity:Foo" does not exist'
         );
 
         $this->cmf->getMetadataFor('prefix:ChildEntity:Foo');
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,14 @@ awareness about deprecated code.
 
 # Upgrade to 2.5
 
+## Deprecated `MappingException`
+
+Referring to `Doctrine\Persistence\Mapping\MappingException` with `catch` or
+`instanceof` is deprecated in favor of
+`Doctrine\Persistence\Mapping\Exception\MappingException`.
+
+All methods inside this class are deprecated in favor of standalone exceptions.
+
 ## Deprecated `OnClearEventArgs::clearsAllEntities()` and `OnClearEventArgs::getEntityClass()`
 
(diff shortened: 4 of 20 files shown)


greg0ire:
@derrabus @malarzm if you're OK with this, let's create 2.6.x and retarget to it.

malarzm:
I'm 👎 for 3 reasons, but open to discussion:

1. "make the MappingException class an interface in 3.0" - does it literally mean what was class in 2.x will become an interface? That'll make it pretty hard to support both 2.x and 3.x at the same time, which we wanted to do in ODM
2. ODM has around 50 named constructors in its [MappingException](https://github.com/doctrine/mongodb-odm/blob/2.3.x/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php), while ORM has even more. To be consistent we should split all exceptions into smaller ones and that will be mostly pointless.
3. I believe there is little one can do with mapping exception in the runtime: mapping exception are a "configuration is borked, please fix" and should be a showstopper

FWIW we have introduced XSD validation in ODM (https://github.com/doctrine/mongodb-odm/pull/1814/files) few years ago and nobody complained they don't have any specialized exception to catch :)

src/Persistence/Mapping/Exception/MappingException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Doctrine\Persistence\Mapping\Exception;
+
+use Doctrine\Persistence\Mapping\MappingException as LegacyMappingException;
+
+/**
+ * Will become an interface in 3.0.
```
stof, line 8: if you make a new name, why not making it an interface directly ? Then, each exception class can extend the legacy class and implement the new interface.

stof:
I tend to agree with @malarzm about the fact that projects should not need to catch MappingException individually (or even at all), as they should fix their mapping.

src/Persistence/Mapping/Exception/MappingException.php
greg0ire: Ah yeah, that would be better indeed, I think, I should probably have done it that way for the ORM.

src/Persistence/Mapping/Exception/ClassNotFoundInNamespaces.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Doctrine\Persistence\Mapping\Exception;
+
+use function implode;
+use function sprintf;
+
+final class ClassNotFoundInNamespaces extends MappingException
+{
+    /**
+     * @param class-string $className
```
derrabus, line 13:
```suggestion
```

I think, this exception can safely be raised with an invalid class name.

greg0ire:
@malarzm I didn't realise about 2. and your other arguments are compelling as well.
If I wanted to introduce an interface to address https://github.com/doctrine/orm/pull/7199#issuecomment-1097249126, any idea how to name it?

tests/Persistence/Mapping/ClassMetadataFactoryTest.php
```diff
@@ -136,7 +136,7 @@ public function testGetInvalidAliasedMetadata(): void
 
         $this->expectException(MappingException::class);
```
derrabus, line 137: Shall we switch the expectation to the new class/interface already?

src/Persistence/Mapping/Exception/MappingException.php
derrabus: I agree that we should make this class an interface already. This would allow ORM 2.x to implement it already.

derrabus (changes requested):

derrabus:
> "make the MappingException class an interface in 3.0" - does it literally mean what was class in 2.x will become an interface? That'll make it pretty hard to support both 2.x and 3.x at the same time, which we wanted to do in ODM

If we made the new `MappingException` an interface already as @stof suggested, you can implement compatibility for `^2.6 || ^3`: Throw the new exceptions, catch the interface.

> ODM has around 50 named constructors in its [MappingException](https://github.com/doctrine/mongodb-odm/blob/2.3.x/lib/Doctrine/ODM/MongoDB/Mapping/MappingException.php), while ORM has even more. To be consistent we should split all exceptions into smaller ones and that will be mostly pointless.

I don't think that you need to follow the same pattern. If the ODM keeps its own `MappingException` with lots of named constructors, I don't see a reason why you should not keep it that way.

> I believe there is little one can do with mapping exception in the runtime: mapping exception are a "configuration is borked, please fix" and should be a showstopper

`MappingException` represents a class of error that can occur due to broken mapping. The existence of this class allows us to document which operations can fail because of that and which cannot. This is about defining a clean abstraction and not about giving userland code the oppertunity to "heal" broken configuration.

stof:
> `MappingException` represents a class of error that can occur due to broken mapping. The existence of this class allows us to document which operations can fail because of that and which cannot. This is about defining a clean abstraction and not about giving userland code the oppertunity to "heal" broken configuration.

We're not saying that MappingException should not exist. We are saying that we don't need separate exception types for each kind of broken configuration.

derrabus:
> We're not saying that MappingException should not exist. We are saying that we don't need separate exception types for each kind of broken configuration.

I agree. The primary goal of this PR was to make `MappingException` an interface. Splitting the current `MappingException` was not necessary to achieve that.

malarzm:
> If I wanted to introduce an interface to address https://github.com/doctrine/orm/pull/7199#issuecomment-1097249126, any idea how to name it?

I don't fully follow why we need the interface if the base class is to stay :)

derrabus:
> I don't fully follow why we need the interface if the base class is to stay :)

That would remove the need for a persistence implementation to extend a base class, see https://github.com/doctrine/orm/pull/7199#issuecomment-1095568780

PR #281, opened 2022-04-14
Emit deprecation for LifecycleEventArgs::getEntity

malarzm:
The method itself had `@deprecated` on it for a long time

### DIFF
--- a/src/Persistence/Event/LifecycleEventArgs.php
+++ b/src/Persistence/Event/LifecycleEventArgs.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Persistence\Event;
 
 use Doctrine\Common\EventArgs;
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\Persistence\ObjectManager;
 
 /**
@@ -41,6 +42,13 @@ class LifecycleEventArgs extends EventArgs
      */
     public function getEntity()
     {
+        Deprecation::trigger(
+            'doctrine/persistence',
+            'https://github.com/doctrine/common/pull/222',
+            '%s is deprecated and will be removed in 3.0, use getObject() instead',
+            __METHOD__
+        );
+
         return $this->object;
     }
 


src/Persistence/Event/LifecycleEventArgs.php
```diff
@@ -41,6 +42,13 @@ public function __construct($object, ObjectManager $objectManager)
      */
     public function getEntity()
     {
+        Deprecation::trigger(
+            'doctrine/persistence',
+            'https://github.com/doctrine/persistence/pull/281',
```
greg0ire: Let's refer to the original PR or issue instead? It was before PRs were used, but we still could use the link to the commit I suppose: aca6f031881f71e79f081ebd2a5ae7913d731d33 It does not explain much, so I let you judge what is best.

malarzm: I wanted to, but simple annotation led me to code migration 🙈 I'll dig further

malarzm: yeaaah, deprecated for 10 years now! https://github.com/doctrine/persistence/commit/aca6f031881f71e79f081ebd2a5ae7913d731d33

greg0ire (dismissed):

src/Persistence/Event/LifecycleEventArgs.php
stof: The PR is https://github.com/doctrine/common/pull/222

greg0ire: Oh right, there was a split!

greg0ire (approved):

greg0ire:
Thanks @malarzm !

Ocramius:
This really didn't need runtime behaviour additions: it was fully statically verifiable.

malarzm:
It'd be weird that we emit deprecations for everything but this one tho :)

Ocramius:
Yikes, it's gone awfully sideways then 😱

greg0ire:
@Ocramius what's so wrong about having the deprecation reported through both mechanisms? https://www.doctrine-project.org/policies/deprecation.html clearly needs a refresh, so now would be a good time to have a discussion about this.

PR #284, opened 2022-04-15
More specific parameter type for ClassMetadata::getIdentifierValues

malarzm:
Found this one while fixing ODM

### DIFF
--- a/src/Persistence/Mapping/ClassMetadata.php
+++ b/src/Persistence/Mapping/ClassMetadata.php
@@ -134,6 +134,7 @@ interface ClassMetadata
      *
      * Has to return an empty array if no identifier isset.
      *
+     * @psalm-param T $object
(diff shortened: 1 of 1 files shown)


malarzm:
@franmomu mind taking a look? I'm not sure I fully understand https://github.com/doctrine/persistence/pull/173 's reasoning

stof:
Well, that new usage of the template indeed means that it is not covariant anymore.

franmomu:
Hey @malarzm sorry I've been some days AFK, I'll try to catch up in the following days. It was created as covariant so when using `ClassMetadata<object>` to specify the type of a ClassMetadata object, it would accept also any kind of `ClassMetadata`, like `ClassMetadata<Foo>`.

But using the template as parameter means that it cannot be used as covariant. 🤔

PR #286, opened 2022-04-20
Handle anonymous classes more gracefully

### LINKED ISSUE #24: Registry::getManagerForClass doesn't work with anonymous classes

greg0ire:
Fixes #24

### DIFF
--- a/src/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/src/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -9,6 +9,7 @@ use Doctrine\Deprecations\Deprecation;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Persistence\Proxy;
 use Psr\Cache\CacheItemPoolInterface;
+use ReflectionClass;
 use ReflectionException;
 
 use function array_combine;
@@ -17,6 +18,7 @@ use function array_map;
 use function array_reverse;
 use function array_unshift;
 use function assert;
+use function class_exists;
 use function explode;
 use function is_array;
 use function str_replace;
@@ -222,6 +224,10 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
     {
         if (isset($this->loadedMetadata[$className])) {
             return $this->loadedMetadata[$className];
+        }
+
+        if (class_exists($className, false) && (new ReflectionClass($className))->isAnonymous()) {
+            throw MappingException::classIsAnonymous($className);
         }
 
         // Check for namespace alias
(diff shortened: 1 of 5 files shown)


stof:
I think you also need to update `isTransient`, to avoid making it trigger a deprecation and call `getFqcnFromAlias` if the anonymous class name contains a `:`.

src/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -462,7 +468,7 @@ public function isTransient($className)
         }
 
         // Check for namespace alias
-        if (strpos($className, ':') !== false) {
+        if (! class_exists($className) && strpos($className, ':') !== false) {
```
stof: I would use `! class_exists($className, false)`, to avoid triggering autoloaders for namespace aliases. Excluding anonymous classes does not need to trigger the autoloader.

stof: btw, other `class_exists` checks added in the PR in the anonymous class checks should probably do the same, as they are added *before* the resolution of the real class name which resolves aliases.

greg0ire: I think that makes sense, since we want to test if the class is anonymous and anonymous classes surely don't require any autoloading.

src/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -220,6 +222,10 @@ abstract protected function isEntity(ClassMetadata $class);
      */
     public function getMetadataFor($className)
     {
+        if (class_exists($className, false) && (new ReflectionClass($className))->isAnonymous()) {
```
stof: should we put that after the check of the cache, to avoid paying the cost of this reflection check when retrieving metadata that was already loaded (which is something that the ORM does a lot, including in its critical path) ?

greg0ire: Sure 👍

stof (approved):

PR #294, opened 2022-05-04
 Fix AbstractManagerRegistry::getManagerForClass when proxy class is used.

### LINKED ISSUE #293: Issue with getEntityManagerForClass in 3.*
See the history and reproducer here: https://github.com/symfony/symfony/issues/46219

Seems there is an issue with `getEntityManagerForClass()` and version 3 of this package.

philiprabbett:
Fixes: https://github.com/doctrine/persistence/issues/293

This PR passes the correct class name to the Metadata Factory `isTransient` method whether the original class is a proxy class or a real class. Whenever a proxy class is the original class, `isTransient` will get the parent class name and whenever a real class is the original class, `isTransient` will get the real class name.

### DIFF
--- a/src/Persistence/AbstractManagerRegistry.php
+++ b/src/Persistence/AbstractManagerRegistry.php
@@ -173,6 +173,8 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
             return null;
         }
 
+        $className = $proxyClass->getName();
+
         if ($proxyClass->implementsInterface($this->proxyInterfaceName)) {
             $parentClass = $proxyClass->getParentClass();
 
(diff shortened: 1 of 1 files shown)


src/Persistence/AbstractManagerRegistry.php
```diff
@@ -173,6 +173,8 @@ public function getManagerForClass(string $class)
             return null;
         }
 
+        $className = $proxyClass->getName();
```
malarzm, line 176: Alternatively line 183 (or 185) could be changed to `$class = $parentClass->getName();`. Both solutions are OK :)

malarzm (commented):
Thanks a lot for the PR! Could you please add a test to prevent regression in the future, like the one we're facing now?

src/Persistence/AbstractManagerRegistry.php
stof: I would change the line 183

malarzm:
I took the liberty to prepare #296 as we want to release 3.0.2 soon. Thank you for finding the culprit @philiprabbett!

PR #301, opened 2022-07-04
fix: Normalize class names #300

### LINKED ISSUE #300: Prepended backslash in class name reloads the full metadata from scratch
### Bug Report

Follow up of https://github.com/symfony/symfony/issues/46837

#### Summary

When extracting class metadata, doctrine copies the original class data and caches it only the first time
But following that copy, calls to `addEntityListener` are made which are not repercuted on the alias https://github.com/doctrine/persistence/blob/25ec98a8cbd1f850e60fdb62c0ef77c162da8704/src/Persistence/Mapping/AbstractClassMetadataFactory.php?rgh-link-date=2022-07-04T14%3A09%3A36Z#L180

#### Solution

~~The alias should not be copied, but should just be added to an alias map to retrieve the original cached class straight away~~

Tofandel:
Fixes #300

### DIFF
--- a/psalm.phpstub
+++ b/psalm.phpstub
@@ -1,0 +1,9 @@
+<?php
+/**
+ * @psalm-pure
+ *
+ * @return ($string is class-string ? ($characters is '\\' ? class-string : string) : string)
+ *
+ * @psalm-flow ($string) -> return
+ */
+function ltrim(string $string, string $characters = " \t\n\r\0\x0B") : string {}
--- a/psalm.phpstub
+++ b/psalm.phpstub
@@ -1,5 +1,8 @@
 <?php
 /**
+ * Workaround for psalm to return a class-string when used with a class string
+ * Can be removed when https://github.com/vimeo/psalm/pull/8219 is released
+ *
  * @psalm-pure
  *
  * @return ($string is class-string ? ($characters is '\\' ? class-string : string) : string)
--- a/src/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/src/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -156,7 +156,7 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
     /**
      * Removes the prepended backslash of a class string to conform with how php outputs class names
      *
-     * @psalm-param string $className
+     * @psalm-param class-string $className
      *
      * @psalm-return class-string
      *
@@ -261,6 +261,7 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      *
      * NOTE: This is only useful in very special cases, like when generating proxy classes.
      *
+     * @psalm-param class-string $className
      * @psalm-param CMTemplate $class
      *
      * @return void
--- a/src/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/src/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -159,14 +159,11 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
      * @psalm-param class-string $className
      *
      * @psalm-return class-string
-     *
-     * @psalm-suppress MoreSpecificReturnType
      */
     private function normalizeClassName(string $className): string
     {
         /**
          * @phpstan-ignore-next-line
-         * @psalm-suppress LessSpecificReturnStatement
          */
         return ltrim($className, '\\');
     }
--- a/src/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/src/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -10,6 +10,7 @@ use Psr\Cache\CacheItemPoolInterface;
 use ReflectionClass;
 use ReflectionException;
 
+use function ltrim;
 use function array_combine;
 use function array_keys;
 use function array_map;
@@ -167,7 +168,7 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
          */
         return ltrim($className, '\\');
     }
-    
+
     /**
      * {@inheritDoc}
      *
@@ -177,7 +178,7 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
     public function getMetadataFor(string $className)
     {
         $className = $this->normalizeClassName($className);
-        
+
         if (isset($this->loadedMetadata[$className])) {
             return $this->loadedMetadata[$className];
         }
@@ -250,6 +251,7 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
     public function hasMetadataFor(string $className)
     {
         $className = $this->normalizeClassName($className);
+
         return isset($this->loadedMetadata[$className]);
     }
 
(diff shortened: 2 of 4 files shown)


src/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -153,6 +153,14 @@ abstract protected function initializeReflection(
      */
     abstract protected function isEntity(ClassMetadata $class);
 
+    /**
+     * Removes the prepended backslash of a class string to conform with how php outputs class names
+     */
+    protected function normalizeClassName(string $className): string
```
greg0ire: This method should be `private` or at least `final` if you think it should be used from the children classes.

src/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -153,6 +153,24 @@ abstract protected function initializeReflection(
      */
     abstract protected function isEntity(ClassMetadata $class);
 
+    /**
+     * Removes the prepended backslash of a class string to conform with how php outputs class names
+     *
+     * @psalm-param string $className
```
greg0ire: Isn't the input supposed to be a `class-string` as well?

src/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -153,6 +153,24 @@ abstract protected function initializeReflection(
...
+     *
+     * @psalm-return class-string
+     *
+     * @psalm-suppress MoreSpecificReturnType
```
greg0ire: We centralize that kind of stuff here: https://github.com/doctrine/persistence/blob/3.0.x/psalm.xml

tests/Persistence/Mapping/AbstractClassMetadataFactoryTest.php
```diff
@@ -78,6 +78,22 @@ public function testItThrowsWhenAttemptingToCheckTransientForShortAlias(): void
...
     }
+
+    public function testItGetsTheSameMetadataForBackslashedClassName(): void
+    {
+        $cmf = $this->getMockForAbstractClass(AbstractClassMetadataFactory::class);
+        $cmf
+            ->method('newClassMetadataInstance')
+            ->with(SomeOtherEntity::class)
+            ->willReturn(
+                $this->createMock(ClassMetadata::class)
```
greg0ire:
```suggestion
                $this->createStub(ClassMetadata::class)
```

greg0ire:
If you are having static analysis issues, consider cloning this for Psalm: https://github.com/phpstan/phpstan/issues/7483

src/Persistence/Mapping/AbstractClassMetadataFactory.php
Tofandel: Isn't that for global/file suppression only? I don't see how to suppress only for a method in that file, I wouldn't want to suppress this for the rest of the file as this is just a false negative for this particular method and this warning is usually useful

src/Persistence/Mapping/AbstractClassMetadataFactory.php
Tofandel: Indeed

src/Persistence/Mapping/AbstractClassMetadataFactory.php
Tofandel: I guess I could try adding a stub file as a workaround since I managed to fix this issue in psalm, the stub file would just need this

```
/**
 * @psalm-pure
 *
 * @return ($string is class-string ? ($characters is '\\' ? class-string : string) : string)
 *
 * @psalm-flow ($string) -> return
 */
function ltrim(string $string, string $characters = " \t\n\r\0\x0B") : string {}
```

Tofandel: Can be removed if https://github.com/vimeo/psalm/pull/8219 is accepted

psalm.phpstub
```diff
@@ -0,0 +1,9 @@
+<?php
+/**
+ * @psalm-pure
+ *
+ * @return ($string is class-string ? ($characters is '\\' ? class-string : string) : string)
+ *
+ * @psalm-flow ($string) -> return
+ */
+function ltrim(string $string, string $characters = " \t\n\r\0\x0B") : string {}
```
greg0ire: Please configure your IDE to always add a newline at end of file, this will make the red pictogram disappear.

psalm.phpstub
```diff
@@ -0,0 +1,9 @@
+<?php
+/**
+ * @psalm-pure
```
greg0ire, line 6: Can you please add a link to your PR in that docblock? :pray: 

Thanks a lot by the way, this is awesome, it will fix the issue for many Doctrine packages :)

psalm.phpstub
Tofandel: It would actually be better to add a `.editorconfig` with the project coding standards so that it's automatic, otherwise I'd have to change them for each project ;)
https://github.com/TYPO3/coding-standards/blob/main/.editorconfig

It would cover trailing whitespaces in blank lines, end of file line return, indent size etc etc

Quite an important file to have for a project with multiple people contributing

Tofandel: Want me to add it to this PR?

psalm.phpstub
Tofandel: Done, no problem happy to help 😊

psalm.phpstub
greg0ire: I think it could be a good idea, but please make it a separate PR because I feel it's going to be controversial.

Tofandel: Okay, then this PR should be good to go

greg0ire (approved):

greg0ire:
Note for maintainers or @Tofandel : I think the commits can be squashed.

SenseException (approved):

greg0ire:
Thanks @Tofandel !

writ3it:
Hi,

this PR breaks test ClassMetadataFactoryTest::testHasGetMetadataNamespaceSeparatorIsNotNormalized in doctrine/orm:2.12.x. 

You can check test results here: https://github.com/doctrine/orm/runs/7698594303?check_suite_focus=true.
Test was implemented here: https://github.com/doctrine/orm/commit/884131e51fa7f14fd30aaed0e0d226dcbf79c51b to solve issue described here: https://github.com/doctrine/orm/issues/1750. 

@Tofandel could you check this case? What we can do?

Tofandel:
The threads are very hard to follow, but to me it seems, as per [the last comment](https://github.com/doctrine/orm/issues/1750#issuecomment-162355602), that the linked issue had nothing to do with class normalization, and was related to namespace and a different fix was applied, so it's safe to modify this test to isSame and isTrue instead of isNotSame and isFalse, because the test mentionned was a revert and does nothing to check the issue (unless I didn't understand it in the first place), and only checks for the opposite of what is expected, since `\DoctrineGlobal_Article` is in fact the same class as `DoctrineGlobal_Article`, which is an issue

You can even see that the real tests for the issues are marked as 
```
    /**
     * @group DDC-115
     */
```
But this specific test is not

And that this test was first created in https://github.com/doctrine/orm/commit/120e6942e83c3cfb4adc7798ddfc6bdd73554bc5
And then the next commit just modify this test to check for the exact opposite, which is not the desired behavior, but at the time it was for "performance concerns"

Given that compared to php 5, php 8 is ~120% faster and that doctrine has a lot more caching now and that this really is such a fast triming operation, it's not a valid argument anymore

writ3it:
Thank you a lot for the explanation and improvement!

PR #304, opened 2022-08-06
fix: Backport Normalize class names #300 (#301)

Tofandel:
Backport class name normalization to 2.5.x


malarzm (approved):

malarzm:
Thanks @Tofandel!

writ3it (approved):
psalm is working here too but without stub imo is ok. Thank you!

Tofandel:
@writ3it Yep but there are some tests that don't use strings instead of classes in this version so it makes the psalm class-string a big pain to backport so I just kept it a simple string

PR #306, opened 2022-08-19
Decouple RuntimePublicReflectionProperty::getValue() from Doctrine\Common\Proxy\Proxy

nicolas-grekas:
On the path to decoupling from doctrine/common
Somewhat related to #299

PR #307, opened 2022-10-03
Add Proxy::__setInitialized() and RuntimeReflectionProperty

nicolas-grekas:
On the path to deprecate `Doctrine\Common\Proxy\Proxy`, I figured out that we're going to need adding `__setInitialized()` to the `Proxy` interface. This is because we need it in `RuntimeReflectionProperty` to disable lazy-loading while calling `ReflectionProperty::setValue()`.

The new `RuntimeReflectionProperty` allows getting+setting *any* properties on objects while preventing lazy-loading initialization, which is something that's going to be needed also to move to more capable proxy implementations.

`RuntimePublicReflectionProperty` is not enough and is thus deprecated.

### DIFF
--- a/src/Persistence/Reflection/RuntimeReflectionProperty.php
+++ b/src/Persistence/Reflection/RuntimeReflectionProperty.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Persistence\Reflection;
+
+use Doctrine\Common\Proxy\Proxy as CommonProxy;
+use Doctrine\Persistence\Proxy;
+use ReflectionProperty;
+use ReturnTypeWillChange;
+
+use function method_exists;
+
+/**
+ * PHP Runtime Reflection Property.
+ *
+ * Avoids triggering lazy loading if the provided object
+ * is a {@see \Doctrine\Persistence\Proxy}.
+ */
+class RuntimeReflectionProperty extends ReflectionProperty
+{
+    /** @var string */
+    private $key;
+
+    public function __construct(string $class, string $name)
+    {
+        parent::__construct($class, $name);
+        $this->key = $this->isPrivate() ? "\0" . $class . "\0" . $name : ($this->isProtected() ? "\0*\0" . $name : $name);
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return mixed
+     */
+    #[ReturnTypeWillChange]
+    public function getValue($object = null)
+    {
+        if ($object === null) {
+            return parent::getValue($object);
+        }
+
+        return ((array) $object)[$this->key] ?? null;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @param object|null $object
+     * @param mixed       $value
+     *
+     * @return void
+     */
+    #[ReturnTypeWillChange]
+    public function setValue($object, $value = null)
+    {
+        if (! ($object instanceof Proxy && ! $object->__isInitialized())) {
+            parent::setValue($object, $value);
+
+            return;
+        }
+
+        if ($object instanceof CommonProxy) {
+            $originalInitializer = $object->__getInitializer();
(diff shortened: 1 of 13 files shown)


greg0ire (dismissed):

src/Persistence/Reflection/RuntimeReflectionProperty.php
```diff
@@ -0,0 +1,74 @@
...
+    public function getValue($object = null)
+    {
+        if ($object === null) {
+            return parent::getValue($object);
+        }
+
+        $name = $this->getName();
+        $key  = $this->isPrivate() ? "\0" . $this->class . "\0" . $name : ($this->isProtected() ? "\0*\0" . $name : $name);
+
+        return ((array) $object)[$key] ?? null;
```
beberlei: Did you do any performance or memory analysis on this compared to before? I am a bit scared :)

I never thought about the performance impact of casting an object to array before, maybe its nothing. but this will be executed hundreds to thousands of time every request where the ORM has even moderate use. Even having the getName/isPrivate/isProtected calls here in the hotpath means this change will probably have a 10%/20% hit on hydration.

This change does highlight, that instead of extending reflection we should introduce a "Mapper" abstraction for a cleaner result.

nicolas-grekas: I just did compare several strategies. In my simple test situation:
- `(array)` is the fastest alternative by 25% with one property in the class and slows down when properties are added. It compares positively to the other alternatives up to ~10 properties
- `ReflectionProperty::getValue()` is as fast as a rebound closure.

I interpret this as:
- If we consider that most entities have about ~10 properties, then `(array)` is fine
- It's not obvious to me that a data mapper would provide a significant advantage
- instead, we might want to add an API to get several properties at once - because then `(array)` will always win.

Note also this: the tests of the ORM still pass when removing this method altogether. It looks like we don't need this precaution when accessing properties. Likely because the ORM doesn't try to access properties of uninitialized proxies.

Can we remove the method then?

nicolas-grekas: Note that if this is a concern, it's already in 3.1 since #306 so that this PR doesn't change anything ;)

nicolas-grekas: > Can we remove the method then?

An idea: make using the new `RuntimeReflectionProperty` opt-in via a constructor flag, deprecate not opting in, and remove `RuntimeReflectionProperty::getValue()`.
Please advise :)

beberlei: We did some off the hand so-so benchmarks and we should be fine with this change.

src/Persistence/Reflection/RuntimeReflectionProperty.php
```diff
@@ -0,0 +1,74 @@
...
+        if ($object instanceof CommonProxy) {
+            $originalInitializer = $object->__getInitializer();
+            $object->__setInitializer(null);
+            parent::setValue($object, $value);
+            $object->__setInitializer($originalInitializer);
+
+            return;
+        }
+
+        $object->__setInitialized(true);
```
stof, line 76: Technically, it might not always exist, as the method is only defined in phpdoc in the interface for now, not yet enforced (which is precisely what allows to do that change in a minor version). So I think your removal of the check was a mistake

nicolas-grekas: I removed it because I thought that in practice the method will always be implemented, and removing the check would save a few CPU cycles. But I'm fine adding it back. PR updated.

stof: As the proxy generators are not in doctrine/persistence itself, it would be totally possible for doctrine/persistence to be updated without getting an updated proxy generator defining the new method.

greg0ire (approved):

derrabus (approved):

PR #313, opened 2022-11-11
Change access modifiers of AbstractManagerRegistry properties.

Vitsen15:
I'm integrating the doctrine ORM to our's legacy project, and I need ManagerRegistry instance.
But when I started to implement resetService() abstract method, I found than I cannot get direct access to properties which it should change without reflection. I think that private access modifiers make you use reflection, that is inappropriate in this case, because resetService() implementation should override $connections or $managers arrays of parent class to which child class do not have access.

The conclusion is: private properties that required to be modified in child implementation of abstract method seems to be a bad approach, because it breaks inheritance scope and makes developer to use reflection API.

### DIFF
--- a/src/Persistence/AbstractManagerRegistry.php
+++ b/src/Persistence/AbstractManagerRegistry.php
@@ -20,10 +20,10 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
     private $name;
 
     /** @var string[] */
-    private $connections;
+    protected $connections;
 
     /** @var string[] */
-    private $managers;
+    protected $managers;
 
     /** @var string */
     private $defaultConnection;


derrabus (changes requested):
I'm afraid I haven't fully understood the problem you're trying to solve. Why do you need access to those properties? Can you share a code sample that helps me understand what you're trying to do?

Vitsen15:
@derrabus, when you create own implementation of EntityManager by extending AbstractManagerRegistry you required to implement next 3 methods: getService(), resetService() and getAliasNamespace().

My entity manager now:

```php
class ManagerRegistry extends AbstractManagerRegistry
{
    /**
     * @param $name
     *
     * @return \Doctrine\DBAL\Connection|string
     * @throws \RestApi\Db\Doctrine\Exception\ConnectionNotFoundException
     */
    protected function getService($name)
    {
        if ($name instanceof Connection || $name instanceof EntityManager) {
            return $name;
        }

        $connections = $this->getConnectionNames();

        if (!array_key_exists($name, $connections)) {
            throw new ConnectionNotFoundException();
        }

        return $connections[$name];
    }

    /**
     * @param $name
     *
     * @return void
     */
    protected function resetService($name)
    {
        // TODO: Implement resetService() method.
    }

    /**
     * @param $alias
     *
     * @return string|void
     */
    public function getAliasNamespace($alias)
    {
        // TODO: Implement getAliasNamespace() method.
    }
}
```

How can I properly recreate entity manager if AbstractManagerRegistry::resetManager() uses AbstractManagerRegistry::resetService():

```php
abstract class AbstractManagerRegistry implements ManagerRegistry
{
    /** @var string */
    private $name;

    /** @var string[] */
    private $connections;

    /** @var string[] */
    private $managers;

    /** @var string */
    private $defaultConnection;

    /** @var string */
    private $defaultManager;
...
    /**
     * {@inheritdoc}
     */
    public function resetManager($name = null)
    {
        if ($name === null) {
            $name = $this->defaultManager;
        }
    
        if (! isset($this->managers[$name])) {
            throw new InvalidArgumentException(sprintf('Doctrine %s Manager named "%s" does not exist.', $this->name, $name));
        }
    
        // force the creation of a new document manager
        // if the current one is closed
        $this->resetService($this->managers[$name]);
    
        return $this->getManager($name);
    }
...
}
```

I see that `$this->resetService($this->managers[$name]);` should recreate an EntityManager which then returned. But implementation of resetService() is placed in child class and doesn't have access to hidden parent property `$managers`.

The only one way to reset `$managers[index]` is to use reflection in my ManagerRegistry, but I think that I should have access to it, because I must implement method changing it and using reflection it this case is not appropriate, because it breaks inheritance visibility scope for AbstractManagerRegistry::resetService().

Please correct me if I'm wrong.

derrabus:
But you can call `$this->getManagerNames()` instead of accessing `$this->managers`, right?

Vitsen15:
Right, but there is no setter for `$this->managers` as I see, right?

derrabus:
Correct, that property is considered immutable. Your code sample does not write to that property either does it?

Vitsen15:
Right, because I'm unable to.
So could you explain what purpose of AbstractManagerRegistry::resetManager()?
According `Doctrine\Persistence\ManagerRegistry` interface is:

```
    /**
     * Resets a named object manager.
     *
     * This method is useful when an object manager has been closed
     * because of a rollbacked transaction AND when you think that
     * it makes sense to get a new one to replace the closed one.
     *
     * Be warned that you will get a brand new object manager as
     * the existing one is not useable anymore. This means that any
     * other object with a dependency on this object manager will
     * hold an obsolete reference. You can inject the registry instead
     * to avoid this problem.
     *
     * @param string|null $name The object manager name (null for the default one).
     *
     * @return ObjectManager
     */
    public function resetManager($name = null);
```

And code of `Doctrine\Persistence\AbstractManagerRegistry` also assumes that:
```
    // force the creation of a new document manager
    // if the current one is closed
    $this->resetService($this->managers[$name]);
```

It seems for me that `$this->resetService($this->managers[$name]);` should recreate specified EntityManager and put it to `$this->managers`, right?

How can I reset a named object manager in `$this->managers`, if property is immutable?

derrabus:
> Right, because I'm unable to.

Okay, but that was kind of what I asked you to provide. Without a valid use case that is made impossible because of the current access modifiers, how can I approve this PR?

> It seems for me that `$this->resetService($this->managers[$name]);` should recreate specified EntityManager and put it to `$this->managers`, right?

No, again: why would you need to change the `$managers` array? It contains a mapping `$managerName => $serviceId`. Why do you need to change that mapping?

Vitsen15:
> Without a valid use case that is made impossible because of the current access modifiers, how can I approve this PR?

I don't understand which example do you need? I'm just trying to implement abstract methods as they designed. But design of access modifiers prevents me to do this.
> So could you explain what purpose of AbstractManagerRegistry::resetManager()?


> No, again: why would you need to change the `$managers` array? It contains a mapping `$managerName => $serviceId`. Why do you need to change that mapping?

Okay, if we don't need to change the mapping, could you explain, why `AbstractManagerRegistry::resetService()`exists?

```
    /**
     * Resets the given services.
     *
     * A service in this context is connection or a manager instance.
     *
     * @param string $name The name of the service.
     *
     * @return void
     */
    abstract protected function resetService($name);
```

The service is connection or entity manager as described in phpdoc section of `Doctrine\Persistence\AbstractManagerRegistry above, so how we're going to reset the services if we don't have access to them? This seems to be a kind of architecture omission isn't it?

The main question is: why do we have abstract method that assumed to change some parent class properties when this properties considered to be immutable and cannot be accessed from child class?

Maybe I misunderstand something, but unfortunately there is no any documentation about manager registry and how properly use and implement it, FYI.

derrabus:
> I don't understand which example do you need? I'm just trying to implement abstract methods as they designed. But design of access modifiers prevents me to do this.

Yet you fail to elaborate why you need direct access to those properties. I'm very sorry, but I'm going to close the PR now. This discussion started as a change request, but it shifts more and more in the direction of a support ticket.

My time is limited and I cannot compensate for the missing or incomplete documentation. On the other hand, creating your own Persistence implementation is a pretty uncommon thing to do. Maybe you can find some help on https://github.com/doctrine/orm/discussions.

malarzm:
@Vitsen15 please take a look at how existing implementations are written, for instance Symfony's: https://github.com/symfony/doctrine-bridge/blob/6.1/ManagerRegistry.php. This should be the best example how the class extending `AbstractManagerRegistry` should look like.

PR #314, opened 2022-11-16
Add missing `throws` annotations

drupol:
As discussed yesterday on Slack,

Here's the missing `throws` annotations.

A following PR is incoming for `doctrine/persistence`.

Related PR: https://github.com/doctrine/orm/pull/10235

### DIFF
--- a/src/Persistence/ObjectManager.php
+++ b/src/Persistence/ObjectManager.php
@@ -6,6 +6,7 @@ namespace Doctrine\Persistence;
 
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\ClassMetadataFactory;
+use Exception;
 
 /**
  * Contract for a Doctrine persistence layer ObjectManager class to implement.
@@ -24,6 +25,8 @@ interface ObjectManager
      * @return object|null The found object.
      * @psalm-return T|null
      *
+     * @throws Exception
+     *
      * @template T of object
      */
     public function find(string $className, $id);
@@ -39,6 +42,8 @@ interface ObjectManager
      * @param object $object The instance to make managed and persistent.
      *
      * @return void
+     *
+     * @throws Exception
      */
     public function persist(object $object);
 
@@ -90,6 +95,8 @@ interface ObjectManager
      * database.
      *
      * @return void
+     *
+     * @throws Exception
      */
     public function flush();
 


src/Persistence/ObjectManager.php
```diff
@@ -4,6 +4,9 @@
 
 namespace Doctrine\Persistence;
 
+use Doctrine\DBAL\Exception;
+use Doctrine\ORM\Exception\ORMException;
+use Doctrine\ORM\ORMInvalidArgumentException;
```
greg0ire: You can reference the DBAL or the ORM here; you would have to use an ancestor that is common to all three types, and known by `doctrine/persistence`.

drupol: so I only use 1 type of Exception instead of 3 ?

src/Persistence/ObjectManager.php
```diff
@@ -4,6 +4,7 @@
 
 namespace Doctrine\Persistence;
 
+use Doctrine\DBAL\Exception;
```
greg0ire: You can't have a dependency to the DBAL here. `persistence` does not, and should not depend on DBAL, that's a circular dependency.

drupol: But then, how the static analysis is supposed to find that `Doctrine\DBAL\Exception;` is extending `Throwable` ?

greg0ire: What are you referring to? It extends `\Exception`, so of course it also extends `Throwable`. I'm not only saying `doctrine/dbal` should not appear in `composer.json`, I'm also saying DBAL should not happen in `src`

drupol: This static analysis issue: https://github.com/doctrine/persistence/actions/runs/3478376544/jobs/5815671037

drupol: Ahhh in `src`, ok, going to do the modifications :)

drupol: Job done!

derrabus (changes requested):
Sorry, but I see no point in documenting that an implementation of the Persistence interfaces might throw virtually any exception.

drupol:
> Sorry, but I see no point in documenting that an implementation of the Persistence interfaces might throw virtually any exception.

I agree. But then, what do you suggest then?

derrabus:
Close this?

drupol:
No. The `find()` method might throw an exception, how would you document it then?

derrabus:
Treat those exceptions as unchecked and don't document them.

drupol:
Can you elaborate when you say "treat those exceptions as unchecked" ?

I'm opening this PR because a couple of days ago, I was trying to make a simple query to a database through a `$repository->find()` method and got an exception because I forgot to enable the VPN.

I was surprised to not see this `Exception` documented. What do you suggest for those cases?

derrabus:
The persistence contracts don't define exceptions, so all exceptions are an implementation detail. And we cannot document that on a generic contract.

drupol:
That I understand, but then what happen when the connection to the database is failing, it should `throw`, no ?

If not, it means that calling the method `find` might be completely silent to any database connection issue and thus unreliable?

Therefore, I believe that adding a `@throws` there is not completely useless.

derrabus:
> That I understand, but then what happen when the connection to the database is failing, it should `throw`, no ?

Sure.

> Therefore, I believe that adding a `@throws` there is not completely useless.

Sure, but `@throws \Exception` is pointless. Give me something worth merging and we discuss it. Or open an issue first.

stof:
I would also say that if the DB connection fails, you would want this exception to be handled by your global exception handler, not by a catch block around the `find()` call.

drupol:
> I would also say that if the DB connection fails, you would want this exception to be handled by your global exception handler, not by a catch block around the `find()` call.

How would you do that when you inject a `$repository` in a controller and call the method `find` then ? I'm genuinely curious.

stof:
Well, I would not make each controller responsible to handle what happens when your DB is unavailable. Let this be handled by the global error handling system (for instance, in Symfony, let this be handled by the ErrorHandler component with its error rendering)

drupol:
True, you're right when it comes to Controllers. But sorry to insist, how to deal when injecting a repository in another service? How that service is supposed to know that calling `$repository->find` is maybe going to throw an exception?

stof:
What is true for controllers is true for other services as well. I don't expect those other services to be responsible to catch such errors either (otherwise, we would have to add `@throws Throwable` on **all** methods to account for errors thrown due to such broken states or due to programmer errors).

PR #326, opened 2023-01-07
Copy classUtils from doctrine/common

VincentLanglet:
I thought the suggestion from @stof in https://github.com/doctrine/data-fixtures/pull/419#issuecomment-1372550411 were pretty good.

`doctrine/common` exposes a `ClassUtils` class with multiple useful methods related to Proxies, (especially a `getRealClass` needed in some repositories). Since Proxy are persistence-related, this would make more sens to have this utility method here to help libraries dropping dependencies to doctrine/common.

I just copy pasted the code from doctrine/common. I dunno if some method of ClassUtils should be dropped or if we keep all of them.

cc @greg0ire


src/Persistence/Util/ClassUtils.php
```diff
@@ -0,0 +1,106 @@
+<?php
+
+namespace Doctrine\Persistence\Util;
```
greg0ire: Let's not encourage "utils". Such namespaces and classes usually end up being piles of unrelated functionality. Also, let's copy exactly what you need, nothing more. Let's not create stuff unless we know it's actually needed. This is related to proxies, so put it in a `Proxy` namespace?

src/Persistence/Util/ClassUtils.php
```diff
@@ -0,0 +1,106 @@
...
+
+use Doctrine\Persistence\Proxy;
+use ReflectionClass;
+
+/**
+ * Class and reflection related functionality for objects that
+ * might or not be proxy objects at the moment.
+ */
+class ClassUtils
```
greg0ire: `ProxyConverter`? There has to be a better name…

VincentLanglet: ProxyResolver ?

src/Persistence/Proxy/ProxyResolver.php
```diff
@@ -0,0 +1,123 @@
...
+use function ltrim;
+use function rtrim;
+use function strrpos;
+use function substr;
+
+/**
+ * Class and reflection related functionality for objects that
+ * might or not be proxy objects at the moment.
+ */
+class ProxyResolver
```
greg0ire, line 21: 👍

src/Persistence/Proxy/ProxyResolver.php
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @param string $className
+     * @psalm-param class-string<Proxy<T>>|class-string<T> $className
+     *
+     * @return string
+     * @psalm-return class-string<T>
+     *
+     * @template T of object
+     */
+    public static function getRealClass($className)
```
greg0ire: You can use native type declarations here.

malarzm:
> Since Proxy are persistence-related, this would make more sens to have this utility method here to help libraries dropping dependencies to doctrine/common.

`doctrine/persistence` already provides https://github.com/doctrine/persistence/blob/3.1.x/src/Persistence/Mapping/ProxyClassNameResolver.php that can be implemented in `doctrine/common` (as that's where Doctrine's proxy implementation lives). Then it just needs to be exposed through ORM and ODM (ODM already does https://github.com/doctrine/mongodb-odm/blob/eeedec20c25ae746b57c9881ff011502eed84a87/lib/Doctrine/ODM/MongoDB/DocumentManager.php#L297). Userland code then depends on an interface effectively decoupling users from `doctrine/common` without dragging static methods from one library to another

greg0ire:
I knew that name was familiar… the ODM method is deprecated though, in favor of using a solution that you could use @VincentLanglet  :thinking: 

Instead of reimplementing `getRealClass` inside `ReferenceRepository`, can't you use `$this->objectManager->getMetadataFor($theClass)->getName()`?

VincentLanglet:
> `doctrine/persistence` already provides [`3.1.x`/src/Persistence/Mapping/ProxyClassNameResolver.php](https://github.com/doctrine/persistence/blob/3.1.x/src/Persistence/Mapping/ProxyClassNameResolver.php?rgh-link-date=2023-01-07T12%3A29%3A38Z) that can be implemented in `doctrine/common` (as that's where Doctrine's proxy implementation lives). Then it just needs to be exposed through ORM and ODM (ODM already does [doctrine/mongodb-odm@`eeedec2`/lib/Doctrine/ODM/MongoDB/DocumentManager.php#L297](https://github.com/doctrine/mongodb-odm/blob/eeedec20c25ae746b57c9881ff011502eed84a87/lib/Doctrine/ODM/MongoDB/DocumentManager.php#L297)). Userland code then depends on an interface effectively decoupling users from `doctrine/common` without dragging static methods from one library to another

Looking at the ODM implementation 
https://github.com/doctrine/mongodb-odm/blob/eeedec20c25ae746b57c9881ff011502eed84a87/lib/Doctrine/ODM/MongoDB/Proxy/Resolver/ProxyManagerClassNameResolver.php#L30-L37
it supports `class-string<T>|class-string<ProxyManager\Proxy\ProxyInterface<T>>` which is not the same that required in the persistence interface `class-string<Doctrine\Persistence\Proxy<T>>|class-string<T>`

If we look at the implementation, it rely on https://github.com/Ocramius/ProxyManager/blob/dbcdf2c925c774d3521107174672665a2f0f7255/src/ProxyManager/Inflector/ClassNameInflector.php#L34
which is the same implementation than doctrine/common, but with a different proxyMarker
https://github.com/Ocramius/ProxyManager/blob/dbcdf2c925c774d3521107174672665a2f0f7255/src/ProxyManager/Inflector/ClassNameInflectorInterface.php#L17
than the one in the `Doctrine\Persistence\Proxy`

Isn't this situation weird ?

VincentLanglet:
> Instead of reimplementing `getRealClass` inside `ReferenceRepository`, can't you use `$this->objectManager->getMetadataFor($theClass)->getName()`?

Done in https://github.com/doctrine/data-fixtures/pull/421

I'll close this PR then.

greg0ire:
> Isn't this situation weird ?

I think that's the reason we have the resolver in the first place, isn't it? See https://github.com/doctrine/persistence/pull/145

PR #327, opened 2023-01-18
Improve the ACMF::doLoadMetadata docblock

mpdude:


derrabus:
PHPCS fails. 🙂

mpdude:
Why can’t it just automatically commit the necessary fixes?

derrabus:
🤷🏻‍♂️

greg0ire:
Hey why don't you use a `pre-push` hook instead of blaming our beautiful pipeline? It will be useful for other commands that can't be autofixed, and will be better for the planet in cases where phpcs is not the only failing job:

```bash
#!/bin/bash
set -e
echo ''|vendor/bin/phpcs
vendor/bin/phpstan
vendor/bin/psalm
vendor/bin/phpunit
```

greg0ire (approved):

PR #333, opened 2023-05-16
Support array of enums

lstrojny:
Support array of enums, so that e.g. with ODM you can do something like this

```php
#[Field(enumType: Variant::class]
private array $variants = [];
```

### DIFF
--- a/src/Persistence/Reflection/EnumReflectionProperty.php
+++ b/src/Persistence/Reflection/EnumReflectionProperty.php
@@ -8,6 +8,9 @@ use BackedEnum;
 use ReflectionProperty;
 use ReturnTypeWillChange;
 
+use function array_map;
+use function is_array;
+
 /**
  * PHP Enum Reflection Property - special override for backed enums.
  */
@@ -33,7 +36,7 @@ class EnumReflectionProperty extends ReflectionProperty
      *
      * @param object|null $object
      *
-     * @return int|string|null
+     * @return int|string|int[]|string[]|null
      */
     #[ReturnTypeWillChange]
     public function getValue($object = null)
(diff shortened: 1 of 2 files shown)


malarzm:
@lstrojny thanks for the PR! I took the liberty and retargeted this to 3.2.0 as this is a feature.

I took a quick look into ODM and if I'm not mistaken the changes here in `doctrine/persistence` will on its own allow combining `collection` and `hash` types with `enumType`?

lstrojny:
That is correct. In fact I am using the feature branch with ODM and it just works

src/Persistence/Reflection/EnumReflectionProperty.php
```diff
@@ -60,9 +63,40 @@ public function getValue($object = null)
...
+     *
+     * @return ($enum is BackedEnum ? (string|int) : (string[]|int[]))
+     */
+    private function fromEnum($enum)
+    {
+        if (is_array($enum)) {
+            /** @var string[]|int[] $values */
+            $values = array_map([self::class, 'fromEnum'], $enum);
+
+            return $values;
```
malarzm: Let's avoid recursion in favour of something along

```suggestion
            return array_map(fn(BackedEnum $e) => $e->value, $enum);
```

malarzm: But compatible with PHP 7.2 :(

malarzm (dismissed):
Also please squash your commits afterwards

malarzm:
Thanks! I've PRed https://github.com/doctrine/mongodb-odm/pull/2532 with a test for future us. I'll see if there are any plans for releasing 3.2.0 here.

src/Persistence/Reflection/EnumReflectionProperty.php
lstrojny: Done

malarzm (approved):

malarzm:
Thanks @lstrojny! I'll be releasing 3.2.0 shortly so you may get back to stable branch ;)

PR #334, opened 2023-07-10
Add ObjectManager::isUninitializedObject()

nicolas-grekas:
For 3.3

Adding `ObjectManager::isUninitializedObject()` will allow userland to decouple from the `Proxy` interface:

before:
```php
if ($entity instanceof Proxy && !$entity->__isInitialized()) {
    $entity->__load();
}
```

after:
```php
if ($em->isUninitializedObject($entity)) {
    $em->initializeObject($entity);
}
```

Also, I removed `Proxy::__setInitialized()` as I figured out it shouldn't be exposed as public API and it should remain an internal concern.

### DIFF
--- a/src/Persistence/ObjectManager.php
+++ b/src/Persistence/ObjectManager.php
@@ -9,6 +9,8 @@ use Doctrine\Persistence\Mapping\ClassMetadataFactory;
 
 /**
  * Contract for a Doctrine persistence layer ObjectManager class to implement.
+ *
+ * @method bool isUninitializedObject(mixed $value) Implementing this method will be mandatory in version 4.
  */
 interface ObjectManager
 {
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,12 +6,14 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
-# Upgrade to 3.1
+# Upgrade to 3.3
 
-## Added method `Proxy::__setInitialized()`
+## Added method `ObjectManager::isUninitializedObject()`
 
-Classes implementing `Doctrine\Persistence\Proxy` should implement the new
+Classes implementing `Doctrine\Persistence\ObjectManager` should implement the new
 method. This method will be added to the interface in 4.0.
+
+# Upgrade to 3.1
 
 ## Deprecated `RuntimePublicReflectionProperty`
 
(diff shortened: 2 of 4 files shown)


src/Persistence/ObjectManager.php
```diff
@@ -9,6 +9,8 @@
 
 /**
  * Contract for a Doctrine persistence layer ObjectManager class to implement.
+ *
+ * @method bool isUninitializedEntity(mixed $value)
```
stof: should this be `mixed` or `object` ?

stof (commented):
I don't think any API of `doctrine/persistence` uses the term `Entity` right now, which is a term used only by the ORM project but not by the ODM.

src/Persistence/ObjectManager.php
nicolas-grekas: it's `mixed` because this is commonly used against any values, eg in the UoW

nicolas-grekas:
OK, renamed to `isUninitializedObject` then.

src/Persistence/ObjectManager.php
```diff
@@ -9,6 +9,8 @@
 
 /**
  * Contract for a Doctrine persistence layer ObjectManager class to implement.
+ *
+ * @method bool isUninitializedObject(mixed $value)
```
greg0ire: The previous comment had a description, so let's keep it?

```suggestion
 * @method bool isUninitializedObject(mixed $value) Implementing this method will be mandatory in version 4.
```

nicolas-grekas: updated thanks

greg0ire (dismissed):

SenseException (dismissed):

greg0ire:
> For 3.3

Wrong branch then?

nicolas-grekas:
I just created branch 3.3.x and changed the target of this PR. But I don't have permission to update the default branch on github.

greg0ire:
Having the stable branch as the default branch is the policy _for this repository_ AFAIK, having the next minor as default branch is a policy for the ODM, and maybe a few more, but not this one, so you did well.

greg0ire (approved):

greg0ire:
Thanks @nicolas-grekas !

src/Persistence/ObjectManager.php
derrabus: The object manager manages objects only, so `object $object` would indeed make sense here imho. It would also be consistent with all other methods on this interface.

PR #338, opened 2023-09-27
Metadata caching isn't compatible with prefixed cache keys

adambalint-srg:
The metadata cache generation, and cache saving didn't work if prefixed cache system is used.

The old functionality checked if the cache item's key is in `$classNames` array. But this array generated only with the `getCacheKey` function, and because of this the items of this array doesn't contain the global prefix of the cache keys.

With the fixed solution we're using only the prefixed keys, and cache item will be saved.


stof:
This would require tests to prevent regressions.

adambalint-srg:
> This would require tests to prevent regressions.

You mean unit tests with prefixed cache reader?

greg0ire:
You should write a unit test that fails without your changes.

adambalint-srg:
> You should write a unit test that fails without your changes.

I wrote this test

src/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -205,17 +206,16 @@ public function getMetadataFor(string $className)
...
 
-                    foreach ($this->cache->getItems(array_keys($classNames)) as $item) {
-                        if (! isset($classNames[$item->getKey()])) {
-                            continue;
-                        }
+                    if ($cacheItems instanceof Traversable) {
+                        $cacheItems = iterator_to_array($cacheItems);
+                    }
+
+                    $loadedMetadataItems = array_combine($loadedMetadata, $cacheItems);
```
stof: this logic assumes that `$cacheItems` is in the same order than `$loadedMetadataItems`. However, I don't think PSR-6 makes any guarantee about the order of items in the returned result.
The CacheItemPoolInterface phpdoc only says that the returned iterable is keyed by cache keys.

adambalint-srg: You're right. This led me to an even simpler solution for this code part. Check my newer commit. It iterates over loaded metadata and get item for each, but I think it's not too frequent that $loadedMetadata's count is more than one, so it's maybe not a performance issue.

adambalint-srg:
@greg0ire Could you help me why is it closed? I think the problem is still valid.

PR #341, opened 2023-12-14
Fix bug in namespace check in MappingDriverChain

ttk:
This fix improves the checking if a namespace appears in a className by ensuring that the namespace ends in a `\`.  Prior to this fix, the following would occur:

A className such as `Doctrine\Tests\Persistence\Mapping\DriverChainEntity` would match the namespace: `Doctrine\Tests\Persistence\Map`, which isn't correct.

Other considerations:  
- what if the namespace already ends in a `\`?  is this a valid case? Probably not.
- what if it's a parent namespace.  In the above example, this would be `Doctrine\Tests\Persistence`.  Should this match against the className `Doctrine\Tests\Persistence\Mapping\DriverChainEntity` or not?  The current logic would match, but I'm not sure if it should....

### DIFF
--- a/src/Persistence/Mapping/Driver/MappingDriverChain.php
+++ b/src/Persistence/Mapping/Driver/MappingDriverChain.php
@@ -8,8 +8,9 @@ use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\MappingException;
 
 use function array_keys;
+use function dirname;
 use function spl_object_hash;
-use function strpos;
+use function trim;
 
 /**
  * The DriverChain allows you to add multiple other mapping drivers for
@@ -73,7 +74,7 @@ class MappingDriverChain implements MappingDriver
     public function loadMetadataForClass(string $className, ClassMetadata $metadata)
     {
         foreach ($this->drivers as $namespace => $driver) {
-            if (strpos($className, $namespace . '\\') === 0) {
+            if (self::isMatchingNamespace($className, $namespace)) {
                 $driver->loadMetadataForClass($className, $metadata);
 
                 return;
@@ -105,7 +106,7 @@ class MappingDriverChain implements MappingDriver
             }
 
             foreach ($driverClasses[$oid] as $className) {
-                if (strpos($className, $namespace . '\\') !== 0) {
+                if (! self::isMatchingNamespace($className, $namespace)) {
                     continue;
                 }
 
@@ -128,7 +129,7 @@ class MappingDriverChain implements MappingDriver
     public function isTransient(string $className)
     {
         foreach ($this->drivers as $namespace => $driver) {
-            if (strpos($className, $namespace . '\\') === 0) {
+            if (self::isMatchingNamespace($className, $namespace)) {
                 return $driver->isTransient($className);
             }
         }
@@ -138,5 +139,13 @@ class MappingDriverChain implements MappingDriver
         }
 
         return true;
+    }
+
+    /**
+     * Checks if the given class name matches the namespace.
+     */
+    protected static function isMatchingNamespace(string $className, string $namespace): bool
+    {
+        return dirname($className) === trim($namespace, '\\');
     }
 }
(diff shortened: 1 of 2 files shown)


src/Persistence/Mapping/Driver/MappingDriverChain.php
```diff
@@ -73,7 +73,7 @@ public function getDrivers()
     public function loadMetadataForClass(string $className, ClassMetadata $metadata)
     {
         foreach ($this->drivers as $namespace => $driver) {
-            if (strpos($className, $namespace) === 0) {
+            if (strpos($className, $namespace . '\\') === 0) {
```
malarzm: The `$namespace` can already contain a `\` at the end. It'd be better to ensure there's always a single `\` in the `addDriver` method

malarzm (changes requested):

derrabus:
Please add a test.

ttk:
The namespace matching code has been revised and a new unit test added.  Note that in the current implementation, the class name will only match the namespace if it's a direct parent namespace.  ie.

`A\B\Classname`  matches `A\B` but not `A`.  I think this is ok because otherwise the wrong driver can be selected if a more global namespace appears first in the list.

stof:
This is not good IMO. Drivers currently support organizing entities in sub-namespaces. The chain driver should not restrict that more

tests/Persistence/Mapping/DriverChainTest.php
```diff
@@ -37,7 +37,7 @@ public function testDelegateToMatchingNamespaceDriver(): void
                 ->with(self::equalTo($className))
                 ->willReturn(true);
 
-        $chain->addDriver($driver1, 'Doctrine\Tests\Models\Company');
+        $chain->addDriver($driver1, 'Doctrine\Tests\Persistence\Map');
```
derrabus, line 40: What's the reason for this change?

tests/Persistence/Mapping/DriverChainTest.php
```diff
@@ -63,7 +63,7 @@ public function testGatherAllClassNames(): void
         $driver1 = $this->createMock(MappingDriver::class);
         $driver1->expects(self::once())
                 ->method('getAllClassNames')
-                ->will(self::returnValue(['Doctrine\Tests\Models\Company\Foo']));
+                ->will(self::returnValue(['Doctrine\Tests\Models\Company\Foo', 'Doctrine\Tests\Models\Company2\Foo']));
```
derrabus, line 66: Same question, why did this test need an adjustment?

tests/Persistence/Mapping/DriverChainTest.php
```diff
@@ -147,6 +147,24 @@ public function testDefaultDriverGetAllClassNames(): void
...
+                return parent::isMatchingNamespace($className, $namespace);
+            }
+        };
+
+        $className = DriverChainEntity::class;
+
+        self::assertTrue($mdc::isMatchingNamespace($className, 'Doctrine\Tests\Persistence\Mapping\\'));
+        self::assertTrue($mdc::isMatchingNamespace($className, 'Doctrine\Tests\Persistence\Mapping'));
+        self::assertFalse($mdc::isMatchingNamespace($className, 'Doctrine\Tests\Persistence\Map'));
+        self::assertFalse($mdc::isMatchingNamespace($className, 'Doctrine\Tests\Persistence'));
```
derrabus, line 166: Those tests don't tell me what problem you're attempting to solve. You're merely testing an internal function, an implementation detail.

greg0ire:
No answer… let's close this I guess.

ttk:
This is still a issue.  Feel free to fork the branch and update it as necessary.

In the meantime, I worked around this bug by changing the name of my directory for the mapping:
Before, which didn't work because the `App` mapping would always be selected:
```
        mappings:
            App:
                dir: '%kernel.project_dir%/src/Entity'
                ...
            AppStatic:
                dir: '%kernel.project_dir%/src/EntityStatic'
                ...
```
to this, which worked, because of the unique prefix:

```
        mappings:
            App:
                dir: '%kernel.project_dir%/src/Entity'
                ...
            AppStatic:
                dir: '%kernel.project_dir%/src/StaticEntity'
                ...
```

PR #343, opened 2024-02-27
Add support for enum arguments to setValue()

greg0ire:
The ORM is relying on that feature.
When trying to use `EnumReflectionProperty` instead of the corresponding class, the test suite breaks on this line: https://github.com/doctrine/orm/blob/b187bc85881cc85d36b28e7ed9dbe2071d472e30/src/UnitOfWork.php#L2406

`EnumReflectionProperty` was added https://github.com/doctrine/persistence/pull/248 so that it could reduce duplication with the ORM, but is not used yet.

### DIFF
--- a/src/Persistence/Reflection/EnumReflectionProperty.php
+++ b/src/Persistence/Reflection/EnumReflectionProperty.php
@@ -10,6 +10,7 @@ use ReturnTypeWillChange;
 
 use function array_map;
 use function is_array;
+use function reset;
 
 /**
  * PHP Enum Reflection Property - special override for backed enums.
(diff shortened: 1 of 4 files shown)


src/Persistence/Reflection/EnumReflectionProperty.php
```diff
@@ -86,13 +86,25 @@ private function fromEnum($enum)
     }
 
     /**
-     * @param int|string|int[]|string[] $value
+     * @param int|string|int[]|string[]|BackedEnum|BackedEnum[] $value
      *
      * @return ($value is int|string ? BackedEnum : BackedEnum[])
```
malarzm:
```suggestion
     * @return ($value is int|string|BackedEnum ? BackedEnum : BackedEnum[])
```

This should take down one SA baseline entry

src/Persistence/Reflection/EnumReflectionProperty.php
```diff
@@ -86,13 +86,25 @@ private function fromEnum($enum)
...
      */
     private function toEnum($value)
     {
+        if ($value instanceof BackedEnum) {
+            return $value;
+        }
+
         if (is_array($value)) {
+            foreach ($value as $v) {
```
malarzm: This was hard to read for me (maybe because I'm in vacations-mode ;)... IMO using `reset` would be more elegant than this entire loop

greg0ire: Enjoy your vacation!

src/Persistence/Reflection/EnumReflectionProperty.php
greg0ire: Didn't seem to, but good point anyway.

SenseException (approved):

malarzm (approved):

PR #348, opened 2024-03-01
Add missing proxy methods

greg0ire:
I somehow thought I had a green build for the ORM after #343, but I really didn't.

Picked the patch release since it's something I forgot and the release is fresh, but please do tell if you think I should be strict about this and target 3.4.x instead.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -11,6 +11,16 @@ parameters:
 			path: src/Persistence/Reflection/EnumReflectionProperty.php
 
 		-
+			message: "#^Method Doctrine\\\\Persistence\\\\Reflection\\\\EnumReflectionProperty\\:\\:getDeclaringClass\\(\\) return type with generic class ReflectionClass does not specify its types\\: T$#"
+			count: 1
+			path: src/Persistence/Reflection/EnumReflectionProperty.php
+
+		-
+			message: "#^Method Doctrine\\\\Persistence\\\\Reflection\\\\EnumReflectionProperty\\:\\:getType\\(\\) should return ReflectionNamedType\\|ReflectionUnionType\\|null but returns ReflectionType\\|null\\.$#"
+			count: 1
+			path: src/Persistence/Reflection/EnumReflectionProperty.php
+
+		-
 			message: "#^Method Doctrine\\\\Persistence\\\\Reflection\\\\EnumReflectionProperty\\:\\:toEnum\\(\\) should return array\\<BackedEnum\\>\\|BackedEnum but returns array\\<BackedEnum\\|int\\|string\\>\\.$#"
 			count: 1
 			path: src/Persistence/Reflection/EnumReflectionProperty.php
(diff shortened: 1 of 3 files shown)


phpstan-baseline.neon
```diff
@@ -10,6 +10,16 @@ parameters:
 			count: 1
 			path: src/Persistence/Reflection/EnumReflectionProperty.php
 
+		-
+			message: "#^Method Doctrine\\\\Persistence\\\\Reflection\\\\EnumReflectionProperty\\:\\:getDeclaringClass\\(\\) return type with generic class ReflectionClass does not specify its types\\: T$#"
```
stof, line 14: This should be solved by making EnumReflectionProperty generic on the class name, as done in the stubs for ReflectionProperty in Psalm and phpstan instead of adding it in the baseline IMO.

src/Persistence/Reflection/EnumReflectionProperty.php
```diff
@@ -30,6 +32,30 @@ public function __construct(ReflectionProperty $originalReflectionProperty, stri
...
     }
 
+    /** {@inheritDoc} */
+    public function getDeclaringClass(): ReflectionClass
+    {
+        return $this->originalReflectionProperty->getDeclaringClass();
+    }
+
+    /** {@inheritDoc} */
+    public function getName(): string
```
stof, line 50: you should be adding the `@psalm-immutable` annotation (which is not the same than `@psalm-pure` that you tried)

greg0ire: I tried it locally, to no avail, then noticed I still had Psalm 4, apparently because Psalm 5 is not compatible with php 8.3 yet, so I just tried pushing, and it doesn't work with Psalm 5 either.

greg0ire: https://psalm.dev/r/f5566261cf :thinking:

phpstan-baseline.neon
greg0ire: I agree, but I'm not sure how I can make the type flow from the constructor to there :thinking: I think I'm supposed to leverage `ReflectionProperty`'s template types, if there are some, but I don't know where to find their definitions.

src/Persistence/Reflection/EnumReflectionProperty.php
stof: https://psalm.dev/r/c2cfe02f5a

phpstan-baseline.neon
stof: hmm, apparently, Psalm makes ReflectionProperty generic but phpstan does not. So I'm not sure solving this one is actually possible.

src/Persistence/Reflection/EnumReflectionProperty.php
greg0ire: Just `@psalm-external-mutation-free` seems sufficient.

SenseException (approved):

PR #357, opened 2024-05-16
fix: Failing to retrieve reflection property on enum

Goazil:
This PR should fix an issue with the property isPrivate from inheritance on `ReflectionProperty`.

 In `EnumReflectionProperty` the method `isPrivate` should be executed on the `originalReflectionProperty` to prevent fatal error : 

```
Internal error: Failed to retrieve the reflection object
```

This commit just override the method and check the privacy on the originalReflectionProperty.

### DIFF
--- a/tests_php81/Persistence/Reflection/EnumReflectionPropertyTest.php
+++ b/tests_php81/Persistence/Reflection/EnumReflectionPropertyTest.php
@@ -125,6 +125,12 @@ class EnumReflectionPropertyTest extends TestCase
         $reflProperty = new EnumReflectionProperty(new ReflectionProperty(TypedEnumClass::class, 'suit'), Suit::class);
         self::assertStringContainsString('@MyDoc', $reflProperty->getDocComment());
     }
+
+    public function testIsPrivate(): void
+    {
+        $reflProperty = new EnumReflectionProperty(new ReflectionProperty(TypedEnumClass::class, 'suit'), Suit::class);
+        self::assertFalse($reflProperty->isPrivate());
+    }
 }
 
 #[Attribute(Attribute::TARGET_PROPERTY)]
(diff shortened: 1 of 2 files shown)


greg0ire (commented):
Please cover this with a test in https://github.com/doctrine/persistence/blob/3.3.x/tests_php81/Persistence/Reflection/EnumReflectionPropertyTest.php

Goazil:
> Please cover this with a test in https://github.com/doctrine/persistence/blob/3.3.x/tests_php81/Persistence/Reflection/EnumReflectionPropertyTest.php

Thanks, done.

greg0ire (dismissed):

tests_php81/Persistence/Reflection/EnumReflectionPropertyTest.php
```diff
@@ -125,6 +125,11 @@ public function testGetDocComment(): void
         $reflProperty = new EnumReflectionProperty(new ReflectionProperty(TypedEnumClass::class, 'suit'), Suit::class);
         self::assertStringContainsString('@MyDoc', $reflProperty->getDocComment());
     }
+    public function testIsPrivate(): void
```
greg0ire, line 129: A line break before that function would be nice

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/3.3.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

greg0ire (approved):

derrabus (approved):

greg0ire:
Thanks @Goazil

PR #361, opened 2024-05-21
Drop support for PHP < 8.1

greg0ire:


composer.json
```diff
@@ -31,7 +31,7 @@
         "phpstan/phpstan-strict-rules": "^1.1",
         "doctrine/coding-standard": "^11",
         "doctrine/common": "^3.0",
-        "phpunit/phpunit": "^8.5 || ^9.5",
+        "phpunit/phpunit": "^9.5",
```
derrabus:
```suggestion
        "phpunit/phpunit": "^9.6",
```

src/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -35,30 +35,22 @@
...
-     *
-     * @var string
-     */
-    protected $cacheSalt = '__CLASSMETADATA__';
+    /** Salt used by specific Object Manager implementation. */
+    protected string $cacheSalt = '__CLASSMETADATA__';
 
-    /** @var CacheItemPoolInterface|null */
-    private $cache;
+    private ?CacheItemPoolInterface $cache = null;
```
derrabus: Shouldn't our CS fix this to…

```suggestion
    private CacheItemPoolInterface|null $cache = null;
```

…?

greg0ire: Ah I wrongly assumed that it didn't because we were using an old version, but it must be because of the php version that I neglected to bump in the config file. Hang on.

derrabus: Indeed, we should maybe bump the CS package version on the lower branches first.

greg0ire: Yes, the bulk of the change comes from the php version upgrade though, bumping to v12 does not achieve much: https://github.com/doctrine/persistence/pull/363

SenseException (dismissed):

src/Persistence/Mapping/Driver/FileDriver.php
```diff
@@ -167,13 +149,11 @@ abstract protected function loadMappingFile(string $file);
...
      * caching is used.
-     *
-     * @return void
      */
-    protected function initialize()
+    protected function initialize(): void
     {
         $this->classCache = [];
-        if ($this->globalBasename === null) {
+        if ($this->globalBasename === '') {
```
derrabus, line 156: Is this a bugfix? Do we need to fix this on 3.x?

SenseException: It would make sense to add this to 3.x. I haven't seen a way to set this property to `null`.

greg0ire: Ah yes I did it then forgot about it. See #362

SenseException (approved):

PR #362, opened 2024-05-21
Handle the correct zero-value

greg0ire:
It was changed in 3e2fad57e1f180e9d6f827c17548171959c1a9a7, I am not sure why, ~I think it might be to avoid the extra conditions / complexity I'm adding here.~ What I know is that it makes a guard condition in `FileDriver::initialize()` useless, most likely resulting in persistence looking for a file with a name consisting only in an extension (`.php`, `.yml`…).

https://github.com/doctrine/persistence/blob/267cd9f317ae4fa9d3b2b02726bbf9cb060da3e6/src/Persistence/Mapping/Driver/FileDriver.php#L173-L191

### DIFF
--- a/src/Persistence/Mapping/Driver/FileDriver.php
+++ b/src/Persistence/Mapping/Driver/FileDriver.php
@@ -173,7 +173,7 @@ abstract class FileDriver implements MappingDriver
     protected function initialize()
     {
         $this->classCache = [];
-        if ($this->globalBasename === null) {
+        if ($this->globalBasename === '') {
             return;
         }
 


greg0ire:
Argh… also since 3e2fad57e1f180e9d6f827c17548171959c1a9a7, `FileLocator::getAllClassNames()` wants a string. Then the right fix must be to handle the empty string properly :facepalm: … and maybe to allow null in the next major release :thinking:

SenseException (approved):

derrabus:
Can we produce a test that fails without your change?

greg0ire:
> Can we produce a test that fails without your change?

Not a very meaningful one I'm afraid. It would assert that a file named literally `.php` doesn't get loaded even if it exists. I don't think such a scenario should be tested, and I think this guard condition is here for performance only (it spares a call to `is_file()`).

derrabus (approved):

PR #369, opened 2024-05-22
Implement isUninitializedObject() in ObjectManagerDecorator

greg0ire:
That method will be part of the ObjectManager interface in 4.0.x.

It was introduced in https://github.com/doctrine/persistence/pull/334

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "phpstan/phpstan-strict-rules": "^1.1",
         "doctrine/coding-standard": "^12",
         "doctrine/common": "^3.0",
-        "phpunit/phpunit": "^8.5 || ^9.5",
+        "phpunit/phpunit": "^8.5.38 || ^9.5",
         "symfony/cache": "^4.4 || ^5.4 || ^6.0",
         "vimeo/psalm": "4.30.0 || 5.24.0"
     },
--- a/src/Persistence/ObjectManagerDecorator.php
+++ b/src/Persistence/ObjectManagerDecorator.php
@@ -4,8 +4,13 @@ declare(strict_types=1);
 
 namespace Doctrine\Persistence;
 
+use BadMethodCallException;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\ClassMetadataFactory;
+
+use function get_class;
+use function method_exists;
+use function sprintf;
 
 /**
  * Base class to simplify ObjectManager decorators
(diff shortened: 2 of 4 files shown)


src/Persistence/ObjectManagerDecorator.php
```diff
@@ -82,6 +82,12 @@ public function initializeObject(object $obj)
         $this->wrapped->initializeObject($obj);
     }
 
+    /** @param mixed $value */
+    public function isUninitializedObject($value): bool
+    {
+        return $this->wrapped->isUninitializedObject($value);
```
derrabus, line 110: What if the wrapped object manager does not implement the method?

greg0ire: Oh… right… guess I'll retarget this to 4.0.x then :disappointed:

derrabus: No, 3.4.x is fine, you just need to handle that case.

greg0ire: Then to answer your original question, there will be an `Error: Call to undefined method WrappedObjectManager::isUninitializedObject()`.  Not sure how to handle it better, unless maybe an exception is considered better than an error.

composer.json
```diff
@@ -30,7 +30,7 @@
         "phpstan/phpstan-strict-rules": "^1.1",
         "doctrine/coding-standard": "^12",
         "doctrine/common": "^3.0",
-        "phpunit/phpunit": "^8.5 || ^9.5",
+        "phpunit/phpunit": "^8.5.38 || ^9.5",
```
greg0ire, line 33: old versions of PHPUnit seem to struggle with `@requires` annotation that are included in a single-line comment.

src/Persistence/ObjectManagerDecorator.php
SenseException: I neither have another idea how to handle this. I'd suggest an exception too.

derrabus: Sounds like a plan.

greg0ire: Ok, I'll use a BadMethodCallException with a better error message than the error then.

src/Persistence/ObjectManagerDecorator.php
```diff
@@ -82,6 +87,29 @@ public function initializeObject(object $obj)
...
+    /** @param mixed $value */
+    public function isUninitializedObject($value): bool
+    {
+        if (! method_exists($this->wrapped, 'isUninitializedObject')) {
+            $wrappedClass = get_class($this->wrapped);
+
+            throw new BadMethodCallException(sprintf(
+                <<<'EXCEPTION'
+Context: Trying to call %s
+Problem: The wrapped ObjectManager, an instance of %s does not implement this method.
```
SenseException, line 99: `__METHOD__` does also contain the FQCN. You could shorten the "instance of" Exception text if you like.

SenseException (approved):

src/Persistence/ObjectManagerDecorator.php
greg0ire: Not really, since the FQCN at hand is not the same as the one referred to in the "Problem" part. In the context part, I say we are calling the decorator, in the problem part, I talk about the decorated class.

PR #371, opened 2024-05-25
Replace comment with real method signature

greg0ire:


nicolas-grekas (approved):

src/Persistence/ObjectManager.php
```diff
@@ -122,6 +118,9 @@ public function getMetadataFactory(): ClassMetadataFactory;
      */
     public function initializeObject(object $obj): void;
 
+    /** Helper method to check whether a lazy loading proxy or persistent collection has been initialized. */
+    public function isUninitializedObject(mixed $value): bool;
```
derrabus, line 122: Is this really `mixed` or should this be an object at least?

greg0ire: Really `mixed`, see discussion on slack channel proxy-integration

greg0ire: back at my desktop, here is the link: https://doctrine.slack.com/archives/C045LBBNALQ/p1716360854270869

derrabus: 😕 

Not happy about that, but if that's the consensus, … 🤷‍♂️

PR #372, opened 2024-06-19
Fix broken return types in AbstractManagerRegistry

stof:
Connections are not ObjectManager instances.

### DIFF
--- a/src/Persistence/AbstractManagerRegistry.php
+++ b/src/Persistence/AbstractManagerRegistry.php
@@ -36,9 +36,9 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
      *
      * @param string $name The name of the service.
      *
-     * @return ObjectManager The instance of the given service.
+     * @return object The instance of the given service.
      */
-    abstract protected function getService(string $name): ObjectManager;
+    abstract protected function getService(string $name): object;
 
     /**
      * Resets the given services.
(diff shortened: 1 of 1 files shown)


stof:
and of course, the fact that this class is implemented with a single `getService` method accepting ids of either connections or managers and sharing the logic means that static analysis is not happy with it :cry:

greg0ire:
How did you notice it? I find it rather strange the ORM test suite does not pick that up when I ran it locally against Persistence 4… maybe that method is useless?

stof:
The ORM itself does not use the ManagerRegistry.

The ManagerRegistry is meant to be used by project that might have to deal with multiple instances of an ObjectManager to get the right one. You'll find consumers of this class in DoctrineBundle and in symfony/doctrine-bridge.
this issue would be caught by running https://github.com/doctrine/DoctrineBundle/blob/2.12.x/tests/RegistryTest.php

greg0ire:
Counterpart patch PR: https://github.com/doctrine/persistence/pull/375

stof:
And for how I noticed it, I started looking at the diff between 3.3.x and 4.0.x (I did not look at all of it, but it was one of the first files listed in `src/`)

stof:
Merging https://github.com/doctrine/persistence/pull/375 up to the 4.0.x branch will likely fix the SA failures in this PR as well

greg0ire:
@stof I'm waiting on an approval on https://github.com/doctrine/persistence/pull/374 to do 1 merge up instead of 2

greg0ire:
Closing as I believe I have addressed everything.

PR #373, opened 2024-06-19
Changes necessary for compatibility with ORM 3

greg0ire:


src/Persistence/Mapping/ClassMetadata.php
```diff
@@ -35,7 +35,7 @@ public function getIdentifier(): array;
      *
      * @return ReflectionClass<T>
      */
-    public function getReflectionClass(): ReflectionClass;
+    public function getReflectionClass();
```
greg0ire, line 38: In ORM 3, the return type is `ReflectionClass|null`

src/Persistence/Mapping/Driver/FileDriver.php
```diff
@@ -71,7 +71,7 @@ public function getGlobalBasename(): string
      *
      * @throws MappingException
      */
-    public function getElement(string $className): ClassMetadata
+    public function getElement(string $className)
```
greg0ire, line 74: When running the ORM 3 test suite, line 95 causes lots of issues:

`Doctrine\Persistence\Mapping\Driver\FileDriver::getElement(): Return value must be of type Doctrine\Persistence\Mapping\ClassMetadata, SimpleXMLElement returned`

src/Persistence/Mapping/Driver/FileDriver.php
```diff
@@ -139,7 +139,7 @@ public function getAllClassNames(): array
      * @return ClassMetadata[]
      * @psalm-return array<class-string, ClassMetadata<object>>
      */
-    abstract protected function loadMappingFile(string $file): array;
+    abstract protected function loadMappingFile(string $file);
```
greg0ire, line 142: The type declaration was simply not updated in ORM 3

src/Persistence/Proxy.php
```diff
@@ -26,7 +26,7 @@ interface Proxy
      *
      * Acts as a no-op if already initialized.
      */
-    public function __load(): void;
+    public function __load();
```
greg0ire, line 29: For this one, I get `PHP Fatal error:  Declaration of Doctrine\Tests\Proxies\__CG__\Doctrine\Tests\Models\Cache\Country::__load(): Doctrine\ORM\Proxy\ProxyFactory must be compatible with Doctrine\Persistence\Proxy::__load(): void in /home/greg/dev/doctrine-orm/patch/vendor/symfony/var-exporter/LazyGhostTrait.php on line 81`.

Cc @nicolas-grekas

src/Persistence/Mapping/Driver/FileDriver.php
stof, line 142: does the ORM need to return something else than an array here ?

src/Persistence/Proxy.php
stof: this could be avoided by changing the ORM ProxyFactory to alias the implement a `__load` function forwarding the call to the `initializeLazyObject` method of the LazyGhostTrait (and importing `initializeLazyObject` as private) instead of renaming the trait method itself (which returns `$this`)

stof: and this can be done in ORM 3.x IMO

src/Persistence/Mapping/Driver/FileDriver.php
stof: to me, this looks like invalid tests

src/Persistence/Mapping/Driver/FileDriver.php
greg0ire: No, it's just that the return type declaration was not added in ORM 3 (it was forgotten)

src/Persistence/Mapping/Driver/FileDriver.php
stof: hmm, actually, it might be that the new types don't match the actual usages. The XmlDriver of the orm returns `array<string, SimpleXMLElement>` in `loadMappingFile`, which is why `getElement` also returns a `SimpleXMLElement`.

My guess is that the types added for this method are totally wrong, as it looks like each FileDriver child class is free to choose the format of the data they want to cache per class anyway (as the implementation of the public API `loadMetadataForClass` is left to the child classes).
And this method should probably have been defined as `protected` instead.

src/Persistence/Proxy.php
stof: See https://github.com/doctrine/orm/pull/11511

greg0ire:
Closing now that compatibility seems doable: https://github.com/doctrine/orm/pull/11533

PR #375, opened 2024-06-19
Relax type declaration

greg0ire:
`getService()` does not always return ObjectManager instances


stof (approved):

MatTheCat:
Looks like these assertions make some tests fail on Symfony’s workflows (https://github.com/symfony/symfony/actions/runs/9599668041/job/26473995866?pr=57467#step:7:2678 e.g.).

greg0ire:
@MatTheCat probably some too naive mocking is involved. This has to be fixed on Symfony's end IMO.

MatTheCat:
@greg0ire you were right, thanks :grin:

PR #381, opened 2024-06-23
Remove StaticReflectionService

greg0ire:
Additionally, implementations of ReflectionService may no longer return null for calls to `getClass()`.


greg0ire:
I am tempted to make `ReflectionService` a class, and have `RuntimeReflectionService` extend it and be deprecated in favor of it, or even be removed. Thoughts?

SenseException:
Because `RuntimeReflectionService` is the last class implementing `ReflectionService`? Is the goal to have just one implementation for a `ReflectionService`?

greg0ire:
Yeah, AFAIK it's the sole implementation of `ReflectionService`, so I'm not sure if there is still a point in having `ReflectionService` be an interface.

alcaeus (approved):
LGTM. Converting `ReflectionService` to a class is going to be a tough to manage BC break - I don't think it's worth it.

PR #389, opened 2024-10-22
Maintenance

greg0ire:


src/Persistence/Reflection/EnumReflectionProperty.php
```diff
@@ -32,31 +32,16 @@ public function __construct(ReflectionProperty $originalReflectionProperty, stri
         $this->enumType                   = $enumType;
     }
 
-    /**
-     * {@inheritDoc}
-     *
-     * @psalm-external-mutation-free
```
greg0ire, line 38: PHPStan apparently doesn't understand that annotation, but then again, it does not complain when it is gone.

SenseException: I don't understand why it was used for those getters.

SenseException (approved):

tests/Persistence/Mapping/AbstractClassMetadataFactoryTest.php
```diff
@@ -72,8 +72,8 @@ public function testItThrowsWhenAttemptingToCheckTransientForShortAlias(): void
         $cmf = $this->getMockForAbstractClass(AbstractClassMetadataFactory::class);
         $this->expectException(MappingException::class);
         /**
-         * @psalm-suppress ArgumentTypeCoercion
-         * @psalm-suppress UndefinedClass
+         * @phpstan-suppress ArgumentTypeCoercion
+         * @phpstan-suppress UndefinedClass
```
stof, line 76: Suppressing a psalm error with a phpstan-suppress annotation does not make any sense. `@phpstan-suppress` does not exist (the equivalent feature in phpstan uses the `@phpstan-ignore-*` tags, as used just below)

src/Persistence/Reflection/EnumReflectionProperty.php
stof: phpstan memoizes getters by default unless tagged with `@phpstan-impure`, while psalm does not. This `@psalm-external-mutation-free` is required for Psalm to consider that calling a method does not mutate other things and so it can remember refinements it did for calls.

PR #412, opened 2025-01-01
Update LICENSE, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2006-2015 Doctrine Project
+Copyright (c) 2006-2025 Doctrine Project
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2006-2015 Doctrine Project
+Copyright (c) 2006-2025 Doctrine Project
```
greg0ire, line 1: Let's drop the end date entirely, so that we don't have to do this every year.

```suggestion
Copyright (c) 2006 Doctrine Project
```

greg0ire:
Also, you should target the lowest supported branch: https://github.com/doctrine/persistence/blob/133bb2825572ee59e506c0bb7d6c9658506619a4/.doctrine-project.json#L25-L30

LICENSE
SenseException: This should also apply to the other PRs created for the copyright year

PR #413, opened 2025-01-25
Use spl_object_id() instead of spl_object_hash()

rosier:
Small performance tweak.

### DIFF
--- a/src/Persistence/Mapping/Driver/MappingDriverChain.php
+++ b/src/Persistence/Mapping/Driver/MappingDriverChain.php
@@ -8,7 +8,7 @@ use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\MappingException;
 
 use function array_keys;
-use function spl_object_hash;
+use function spl_object_id;
 use function str_starts_with;
 
 /**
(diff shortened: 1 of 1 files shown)


greg0ire:
Please retarget to 4.1.x

greg0ire (approved):

SenseException:
PHP docs:
> This function returns a unique identifier for the object. The object id is unique for the lifetime of the object. Once the object is destroyed, its id may be reused for other objects. This behavior is similar to spl_object_hash().

The behavior seems to be the same, but what are the performance gains by this, @rosier? Did you try some performance tests

derrabus:
I wonder if we should use an `SplObjectStorage` instead of an array, so we don't need to care about object hashing anymore.

rosier:
> The behavior seems to be the same, but what are the performance gains by this, @rosier? Did you try some performance tests

I used the benchmark from
* https://github.com/rectorphp/rector-src/pull/4876

rosier:
> I wonder if we should use an `SplObjectStorage` instead of an array, so we don't need to care about object hashing anymore.

An array + spl_object_id also outperforms `SplObjectStorage`

https://3v4l.org/CZjfW#v8.4.3
https://3v4l.org/CZjfW#v8.3.16
https://3v4l.org/CZjfW#v8.2.27
https://3v4l.org/CZjfW#v8.1.31

SenseException:
As far as I remember was `SplObjectStorage` dropped in the past because it wasn't as performant as the current array solution.

SenseException (approved):

GromNaN (approved):

greg0ire:
Thanks @rosier !

PR #417, opened 2025-04-03
Implement file regex search for ColocatedMappingDriver

rela589n:
Hello,

This PR adds a flexibility for `ColocatedMappingDriver`, allowing client code to customize files search by specifying an regex (rather than just file extension as previously).

This is very useful for the cases when tests, fixtures, factories are kept in the same namespace as entity (having one namespace for one Aggregate).

For example, if we structure classes for User Aggregate:

```
User/
├── User.php
├── UserFixture.php
└── Actions/
    └── Register/
        ├── RegisterUserCommand.php
        └── RegisterUserTest.php
```

Like here we have `User/` namespace that includes everything possessed by the concept of User (e.g. Entity, Fixture, Command, Test, Controller, maybe some Value-Objects, etc.)

Right now Doctrine mapping is not ready for this kind of architecture, as it would scan all these files and include them, while they don't have to be included, so I would like to fix it with regex. It would not have been a big problem with all these files being included if there were no dev-things there, but they are there, so it will possibly break in prod environment, if we keep them.

Therefore, I introduced new `$fileRegex` property that is used when filtering, and it could be set from the client code. 
Old `$fileExtension` I have deprecated so that we could use `$fileRegex`.

Please, let me know if you don't have objections about this, and what further steps I need to do?

Also, please note that this is my first PR to Doctrine Project, so I would appreciate it if you'll be patient with me.

### DIFF
--- a/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
+++ b/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
@@ -8,9 +8,9 @@ use Doctrine\Persistence\Mapping\MappingException;
 use FilesystemIterator;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
-use RecursiveRegexIterator;
 use ReflectionClass;
 use RegexIterator;
+use SplFileInfo;
 
 use function array_merge;
 use function array_unique;
@@ -169,17 +169,18 @@ trait ColocatedMappingDriver
                 throw MappingException::fileMappingDriversRequireConfiguredDirectoryPath($path);
             }
 
+            /** @var iterable<SplFileInfo> $iterator */
             $iterator = new RegexIterator(
                 new RecursiveIteratorIterator(
                     new RecursiveDirectoryIterator($path, FilesystemIterator::SKIP_DOTS),
                     RecursiveIteratorIterator::LEAVES_ONLY,
                 ),
                 $this->fileRegex,
-                RecursiveRegexIterator::GET_MATCH,
+                RegexIterator::MATCH,
             );
 
             foreach ($iterator as $file) {
-                $sourceFile = $file[0];
+                $sourceFile = $file->getPathname();
 
                 if (preg_match('(^phar:)i', $sourceFile) === 0) {
                     $sourceFile = realpath($sourceFile);
--- a/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
+++ b/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
@@ -43,7 +43,14 @@ trait ColocatedMappingDriver
      */
     protected array $excludePaths = [];
 
-    /** The file extension of mapping documents. */
+    /** The regex used to match mapping files. */
+    protected string $fileRegex = '/^.+\.php$/i';
+
+    /**
+     * The file extension of mapping documents.
+     *
+     * @deprecated Use {@see $fileRegex} instead.
+     */
     protected string $fileExtension = '.php';
 
     /**
@@ -94,16 +101,38 @@ trait ColocatedMappingDriver
         return $this->excludePaths;
     }
 
-    /** Gets the file extension used to look for mapping files under. */
+    /** Gets the file regex used to look for mapping files under. */
+    public function getFileRegex(): string
+    {
+        return $this->fileRegex;
+    }
+
+    /** Sets the file regex used to look for mapping files with. */
+    public function setFileRegex(string $fileRegex): void
+    {
+        $this->fileRegex     = $fileRegex;
+        $this->fileExtension = '';
+    }
+
+    /**
+     * Gets the file extension used to look for mapping files under.
+     *
+     * @deprecated Use {@see getFileRegex()} instead.
+     */
     public function getFileExtension(): string
     {
         return $this->fileExtension;
     }
 
-    /** Sets the file extension used to look for mapping files under. */
+    /**
+     * Sets the file extension used to look for mapping files under.
+     *
+     * @deprecated Use {@see setFileRegex()} instead.
+     */
     public function setFileExtension(string $fileExtension): void
     {
         $this->fileExtension = $fileExtension;
+        $this->fileRegex     = '/^.+' . preg_quote($fileExtension) . '$/i';
     }
 
     /**
@@ -145,7 +174,7 @@ trait ColocatedMappingDriver
                     new RecursiveDirectoryIterator($path, FilesystemIterator::SKIP_DOTS),
                     RecursiveIteratorIterator::LEAVES_ONLY,
                 ),
-                '/^.+' . preg_quote($this->fileExtension) . '$/i',
+                $this->fileRegex,
                 RecursiveRegexIterator::GET_MATCH,
             );
 
--- a/tests/Persistence/Mapping/ColocatedMappingDriverTest.php
+++ b/tests/Persistence/Mapping/ColocatedMappingDriverTest.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\Persistence\Mapping;
 
+use Doctrine\Entity;
+use Doctrine\EntityFixture;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\ColocatedMappingDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
@@ -11,7 +13,7 @@ use Doctrine\TestClass;
 use Generator;
 use PHPUnit\Framework\TestCase;
 
-use function array_values;
+use function sort;
 
 class ColocatedMappingDriverTest extends TestCase
 {
@@ -44,6 +46,18 @@ class ColocatedMappingDriverTest extends TestCase
         ], $driver->getExcludePaths());
     }
 
+    public function testGetSetFileRegex(): void
+    {
+        $driver = $this->createDriver(__DIR__ . '/_files/colocated');
+        self::assertSame('/^.+\.php$/i', $driver->getFileRegex());
+
+        $driver->setFileRegex('/^(?!.*Test\.php$).*\.php$/');
+
+        self::assertSame('/^(?!.*Test\.php$).*\.php$/', $driver->getFileRegex());
+        self::assertSame('', $driver->getFileExtension());
+    }
+
+    /** @deprecated */
     public function testGetSetFileExtension(): void
     {
         $driver = $this->createDriver(__DIR__ . '/_files/colocated');
@@ -52,6 +66,7 @@ class ColocatedMappingDriverTest extends TestCase
         $driver->setFileExtension('.php1');
 
         self::assertSame('.php1', $driver->getFileExtension());
+        self::assertSame('/^.+\.php1$/i', $driver->getFileRegex());
     }
 
     /** @dataProvider pathProvider */
@@ -61,7 +76,22 @@ class ColocatedMappingDriverTest extends TestCase
 
         $classes = $driver->getAllClassNames();
 
-        self::assertSame([TestClass::class], $classes);
+        sort($classes);
+        self::assertSame([Entity::class, EntityFixture::class], $classes);
+    }
+
+    public function testGetAllClassNamesWithRegex(): void
+    {
+        $noFixturesRegex = '/^(?!.*Fixture\.php$).*\.php$/';
+        $driver          = $this->createDriver(__DIR__ . '/_files/colocated', $noFixturesRegex);
+
+        $classes = $driver->getAllClassNames();
+
+        self::assertSame(
+            [Entity::class],
+            $classes,
+            'EntityFixture.php should be excluded by the regular expression',
+        );
     }
 
     /** @return Generator<string, array{string}> */
@@ -71,9 +101,9 @@ class ColocatedMappingDriverTest extends TestCase
         yield 'winding path' => [__DIR__ . '/../Mapping/_files/colocated'];
     }
 
-    private function createDriver(string $path): MyDriver
+    private function createDriver(string $path, string|null $fileRegex = null): MyDriver
     {
-        return new MyDriver($path);
+        return new MyDriver([$path], $fileRegex);
     }
 }
 
@@ -81,10 +111,19 @@ final class MyDriver implements MappingDriver
 {
     use ColocatedMappingDriver;
 
-    /** @param string ...$paths One or multiple paths where mapping classes can be found. */
-    public function __construct(string ...$paths)
+    /**
+     * @param non-empty-list<string> $paths     One or multiple paths where mapping classes can be found.
+     * @param string|null            $fileRegex The regex used to look for mapping files with.
+     */
+    public function __construct(array $paths, string|null $fileRegex = null)
     {
-        $this->addPaths(array_values($paths));
+        $this->addPaths($paths);
+
+        if ($fileRegex === null) {
+            return;
+        }
+
+        $this->setFileRegex($fileRegex);
     }
 
     /**
@@ -96,6 +135,6 @@ final class MyDriver implements MappingDriver
 
     public function isTransient(string $className): bool
     {
-        return $className !== TestClass::class;
+        return $className === TestClass::class;
     }
 }
--- a/tests/Persistence/Mapping/ColocatedMappingDriverTest.php
+++ b/tests/Persistence/Mapping/ColocatedMappingDriverTest.php
@@ -4,12 +4,12 @@ declare(strict_types=1);
 
 namespace Doctrine\Tests\Persistence\Mapping;
 
-use Doctrine\Entity;
 use Doctrine\EntityFixture;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\ColocatedMappingDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
-use Doctrine\TestClass;
+use Doctrine\Tests\Persistence\Mapping\_files\colocated\Entity\Entity;
+use Doctrine\Tests\Persistence\Mapping\_files\colocated\Entity\TestClass;
 use Generator;
 use PHPUnit\Framework\TestCase;
 
@@ -77,20 +77,20 @@ class ColocatedMappingDriverTest extends TestCase
         $classes = $driver->getAllClassNames();
 
         sort($classes);
-        self::assertSame([Entity::class, EntityFixture::class], $classes);
+        self::assertSame([EntityFixture::class, Entity::class], $classes);
     }
 
     public function testGetAllClassNamesWithRegex(): void
     {
-        $noFixturesRegex = '/^(?!.*Fixture\.php$).*\.php$/';
-        $driver          = $this->createDriver(__DIR__ . '/_files/colocated', $noFixturesRegex);
+        $entitiesRegex = '/\/Entity\/.+\.php$/';
+        $driver        = $this->createDriver(__DIR__ . '/_files/colocated', $entitiesRegex);
 
         $classes = $driver->getAllClassNames();
 
         self::assertSame(
             [Entity::class],
             $classes,
-            'EntityFixture.php should be excluded by the regular expression',
+            'EntityFixture.php should be excluded by the regex, being not in Entity directory.',
         );
     }
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,25 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 4.1
+
+## Added `ColocatedMappingDriver::$fileRegex`
+
+You can search for mapping files with regex:
+
+```php
+// Search in all Entity/ directories
+$driver->setFileRegex('/\/Entity\/.+\.php$/');
+```
+
+## Deprecated `ColocatedMappingDriver::$fileExtension`
+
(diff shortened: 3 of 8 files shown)


greg0ire:
Hi! I retargeted on 4.1.x since this is no bugfix, but now there are conflicts, please address them.

rela589n:
@greg0ire , fixed the conflicts

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -145,7 +174,7 @@ public function getAllClassNames(): array
                     new RecursiveDirectoryIterator($path, FilesystemIterator::SKIP_DOTS),
                     RecursiveIteratorIterator::LEAVES_ONLY,
                 ),
-                '/^.+' . preg_quote($this->fileExtension) . '$/i',
+                $this->fileRegex,
```
greg0ire, line 178: I think it would be fair to write more tests targeting this change beside testing getters and setters.

greg0ire (commented):
Thanks for your contribution. Which documentations will need to be upgraded if we go with this change?

Also, you should add upgrade instructions in `UPGRADE.md`

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
rela589n: Actually, the resulting regex has never changed. 

So, as it was `'/^.+\.php$/i'`, so it is now:

```php
self::assertSame('/^.+\.php$/i', $driver->getFileRegex());
```

I've added this assertion to `testSetFileRegex` that verifies that regex is actually one we expect same as was before.

greg0ire: And that's great, definitely something we'd want to test. But I think the new functionality you unlocked should be demonstrated in a test based on the filetree you showed in the description of this PR.

rela589n: do you want me to create a new test for it, or maybe I could adjust `testGetAllClassNames()` so that it would use regex to find only needed classes?

stof: I think it should be a new test

rela589n: All right, so I added the test that will collect only needed entities by regex: `testGetAllClassNamesWithRegex`

I think it should be quite obvious that if `testGetAllClassNames` asserts about `[Entity::class, EntityFixture::class]`, while new test asserts only about `[Entity::class]`, this means that fixture has been excluded by regex.

rela589n:
Hi, @stof , @greg0ire , could you check it once more please?

tests/Persistence/Mapping/ColocatedMappingDriverTest.php
```diff
@@ -61,7 +76,18 @@ public function testGetAllClassNames(string $path): void
...
+    }
+
+    public function testGetAllClassNamesWithRegex(): void
+    {
+        $noFixturesRegex = '/^(?!.*Fixture\.php$).*\.php$/';
+        $driver          = $this->createDriver(__DIR__ . '/_files/colocated', $noFixturesRegex);
+
+        $classes = $driver->getAllClassNames();
+
+        self::assertSame([Entity::class], $classes);
```
greg0ire: > I think it should be quite obvious that if testGetAllClassNames asserts about [Entity::class, EntityFixture::class], while new test asserts only about [Entity::class], this means that fixture has been excluded by regex.

You still felt the need to mention it though. Let's make it even more obvious then.

```suggestion
        self::assertSame(
                [Entity::class],
                $classes,
                'EntityFixture.php should be excluded by the regular expression'
        );
```

greg0ire:
> Which documentations will need to be upgraded if we go with this change?

Please answer this.

rela589n:
> >  Which documentations will need to be upgraded if we go with this change?
> 
> Please answer this.

Sorry, I didn't get the point. What documentations do you mean? I thought you were talking about UPGRADE.md, or is it anything else you are talking about?

greg0ire:
I mean: is there a documentation page on doctrine-project.org or maybe even symfony.com detailing how to configure mapping drivers? In the end, how will end-users know how to configure their projects to allow using the file tree you described in your original post?

rela589n:
I see that `ColocatedMappingDriver` is only used in `AttributeDriver`, yet `setFileExtension` is never used.

Doctrine Bundle doesn't set it, as it doesn't have support for it.

Also, I searched a bit through the docs, and see that method `setFileExtension` is all about other places, not related to `ColocatedMappingDriver`:

- `\Doctrine\Common\ClassLoader` ([common](https://www.doctrine-project.org/projects/doctrine-common/en/3.5/reference/class-loading.html))
- `\Doctrine\RST\Configuration` ([rst](https://www.doctrine-project.org/projects/doctrine-rst-parser/en/0.5/customizing-rendering.html))
- `\Doctrine\ORM\Mapping\Driver\XmlDriver` (and `SimplifiedXmlDriver`) ([orm xml driver](https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/xml-mapping.html)) - this one is outdated, since there's no such method as `setFileExtension` for `XmlDriver`, as it's not in abstract `FileDriver`. Right now it seems to be possible with `setLocator`, or `->getLocator()->setFileExtension()` (yet this isn't in `FileLocator` interface), so I think this doc should be updated.

Regarding the new `setFileRegex`, I think it could be referenced in [advanced orm configuration](https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/advanced-configuration.html) options.

rela589n:
@greg0ire , hi, could you please guide on the next steps to take?

greg0ire:
No occurrences in the ODM docs? I think next steps would be to get more approvals on this, so I'll request more reviews. In parallel to that, you could open draft PRs updating the docs, it would help reviewers understand the changes you are proposing, and will be needed anyway.

greg0ire (dismissed):

SenseException (dismissed):

rela589n:
> No occurrences in the ODM docs?

I checked for ODM docs, and found one more entry (relevant to `setFileExtension`, but not to `ColocatedMappingDriver`):
- `Doctrine\ODM\MongoDB\Mapping\Driver\XmlDriver` (and `SimplifiedXmlDriver`) ([odm xml driver](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.9/reference/xml-mapping.html)) - this doc seems to be outdated, since there's no such method as `setFileExtension` in `XmlDriver`, as it's neither in abstract `FileDriver`. This is the same very issue described in my [last comment](https://github.com/doctrine/persistence/pull/417#issuecomment-2791941904), so I think this doc should be updated as well.

No different from already described.

Also, there are few references from Yaml mapping drivers ([odm](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/1.0/reference/yml-mapping.html), [orm](https://www.doctrine-project.org/projects/doctrine-orm/en/2.15/reference/yaml-mapping.html)), though I don't think they should be updated, since they've been removed.

> In parallel to that, you could open draft PRs updating the docs

I'll try to do it soon.

rela589n:
Hi guys, hope you are doing well!

@greg0ire , @SenseException , could you suggest moving forward

greg0ire:
I'll be honest I'm not too crazy about the result (I doubt many developers know about negative lookahead, I'm not even sure I've used one myself ever). I think we should start from the UI, and imo what would be best would be to have the user provide a list of glob patterns for exclusion, and then build a regex from that if necessary. Letting them set a regex is very powerful, but seems error prone and hard to test. Are there other use cases than the one you mentioned for this? If not then maybe the API should be restricted to that use case, and made easier to use.

Also, is there an easy way for users to validate their changes, like a command that allows getting a list of mapping files somewhere?

I'd like to know what you and other maintainers think about this.

rela589n:
You don't have to follow the line of how the green grass eaten by the black cow produced a white milk to enjoy the result, so don't the developers need to know how the negative lookahead works to use it.

> seems error prone and hard to test

It is not. There's an easy tool to test the regex https://regex101.com/ 

This is the most powerful implementation, and it doesn't have any boundaries or limitations whatsoever.

You'd be able to map the entities inside any nested strucutures organized in whatever arbitrary way you might want. You might include mappings from any deep nested `Entity`, or `ValueObject` directories of the project.

Implementing it otherwise would be a deviation from the best option.

> Also, is there an easy way for users to validate their changes, like a command that allows getting a list of mapping files somewhere?

I think you are the one who knows it

greg0ire:
> You don't have to follow the line of how the green grass eaten by the black cow produced a white milk to enjoy the result, so don't the developers need to know how the negative lookahead works to use it.

They don't, what I'm saying is

```yaml
fileRegex: '/^(?!.*Fixture\.php$).*\.php$/'
```

is quite a mouthful compared to

```yaml
excludeGlobs: '**/*Fixture.php'
```

> This is the most powerful implementation, and it doesn't have any boundaries or limitations whatsoever.

Which is why I'm asking what other use cases (than exclusion) it has.

> You'd be able to map the entities inside any nested strucutures organized in whatever arbitrary way you might want. You might include mappings from any deep nested Entity, or ValueObject directories of the project.

Sounds like this could be done with 2 globs, or 2 array of globs.

> Implementing it otherwise would be a deviation from the best option.

It would be a deviation from the simplest option to implement, I'm not sure we can equate that to the best option.

> I think you are the one who knows it

You're assuming I use Doctrine regulary, and that's a fair assumption, but I'm afraid it's very wrong :grimacing:

rela589n:
Do you want to code for every possible solution, already covered by the flexible solution?

I only claim for a flexible solution, not any other.

greg0ire:
Maybe you're right, I'm really not sure about this. I'd like to hear opinions from other maintainers. @beberlei @stof @malarzm @alcaeus @SenseException @derrabus , sorry for the ping, but maybe you have an opinion.

stof:
Asking people to write a single complex regex to include things is quite bad for DX IMO. Such regular expressions will very quickly become unmaintainable (and many devs are not familiar with lookaheads in regex).

I think being able to configure some exclusions (supporting multiple ones to avoid having to write a crazy exclusion rule) would lead to much better DX for the case of excluding a `Fixtures` folder for instance.

greg0ire:
OK, it's not just me. @rela589n I'm truly sorry for not pointing that out earlier.

rela589n:
Why in the world is it bound to the particular scenario of usage?

This PR is not intended to implement exclusion rules. 

It is intended to provide setter for a regex used within

rela589n:
I am as sure about this as I'm sure about God

greg0ire:
Yeah me too that's the problem

derrabus:
I would not want to make the file discovery logic any more complex, tbh. The current simple approach (read all PHP files from `$dir`) works for most of the cases and we should keep this approach as accessible as it is now.

If we want to open this up for more complex cases, we could allow to pass an iterable of files, so people could use Symfony Finder etc. to crawl for files.

rela589n:
Doesn't look like a complication from my side

derrabus:
> Doesn't look like a complication from my side

Shall we cose then? Because I don't think we're getting anywhere from here.

rela589n:
@derrabus, what do you mean keep approach as accessible as it is was?

it remains as accessible as it was

derrabus:
> it remains as accessible as it was

No, you're literally deprecating the current way.

stof:
@rela589n the use case you described to us is the ability to exclude a Fixture folder (or some other files in the folder).
What @greg0ire and I discussed is the fact that there are simpler solutions for that use case, that are a lot easier to use for developers (making developers write a complex regex using lookahead is **not** a simple API)

rela589n:
> No, you're literally deprecating the current way.

What do you mean? Do you mean that I deprecate the ability to search for all files inside a particular directory?
No. It all remains as it was, with only the internals changed (originally `'/^.+' . preg_quote($this->fileExtension) . '$/i',` was part of the method, and now it became a property). Logic has not changed.

rela589n:
@derrabus , slightly updated UPGRADE file; suppose this is more clear now.

derrabus:
> > No, you're literally deprecating the current way.
> 
> What do you mean?

👇🏻 👇🏻 👇🏻 

https://github.com/doctrine/persistence/blob/4a4c43245ec9b1b3065f236e0476c25dd58bdb1a/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php#L130

☝🏻 ☝🏻 ☝🏻

rela589n:
All right, so you mean `setFileExtension` 

Do you think there's any point in setting any other extension different from .php?

In spite of that, "deprecating the current way" doesn't mean file extension (this is not used anyway), but rather "approach of doing things", and the approach remains the same.

derrabus:
> All right, so you mean `setFileExtension`
> 
> Do you think there's any point in setting any other extension different from .php?

Probably not. So your point is that the default behavior (all files with the `.php` extension) is already good enough, so we don't really need a way to set a different extension? Yeah, I can get on board with that.

Then again, what do you think about my proposal to allow a custom iterable with files? I'm afraid that by allowing to change the regex, we open up a slippery slope because the regex solution still might not be flexible enough for someone.

rela589n:
> Then again, what do you think about my proposal to allow a custom iterable with files?

It might be useful, I think. Yet, I would want to have the solution that is configurable without a need to write code.

> I'm afraid that by allowing to change the regex, we open up a slippery slope

Do you think a lot of people are going to use it? I think so. Though, if anybody would like to set the custom regex, they must be conscious of what they are doing.

rela589n:
Hi guys, can we move further?

greg0ire:
> Yet, I would want to have the solution that is configurable without a need to write code.

I think it's not incompatible.

> If we want to open this up for more complex cases, we could allow to pass an iterable of files, so people could use Symfony Finder etc. to crawl for files.

In this case, for instance, maybe something could be contributed to Symfony bundles so that it's possible to call Symfony Finder based on a configuration very similar to what I posted earlier. Does it make sense?

rela589n:
Sorry, but I have no idea of what you are trying to explain to me. 

Right now there's `RegexIterator` iterating over `RecursiveDirectoryIterator` for the given path. 

When I saw that the pattern was not configurable from the outside, I thus decided to make it configurable.

And that's why `$fileRegex` property was ever created.

Of course, you can implement some other scenarios of filtering that I might not be aware of right now, but I don't see what you see

greg0ire:
> Sorry, but I have no idea of what you are trying to explain to me.

OK, let's go step by step. What do you think the following means in term of API changes:

> we could allow to pass an iterable of files

rela589n:
I don't get how you would want to pass the iterable of files.

greg0ire:
I cannot speak for @derrabus , but maybe the suggestion was to introduce a brand new `setFiles(iterable $files): void` method to set a property and then use it in `getAllClassNames()`?

rela589n:
How do you want to correlate it with the existing `$paths` and `$excludePaths`? Are these to be deleted?

It seems a hard job to do, as current `$iterator` is bound both with `$path` and with regex.

greg0ire:
> How do you want to correlate it with the existing $paths and $excludePaths?

I think I would first forbid setting both by throwing an exception. Then we can think about deprecating the old way and eventually remove it.

> It seems a hard job to do, as current $iterator is bound both with $path and with regex.

If an iterable of file is provided, I would skip `$iterator` entirely.

rela589n:
I have checked the code once more, and being honest, I don't want to implement anything else aside from that

greg0ire:
Thanks for your honesty, and that's OK, nobody is forced to do anything. Let us close this then?

rela589n:
Actually, I think regex would be better

rela589n:
@greg0ire , I want this to be merged

greg0ire:
Yes, we all understood that. Did you understand that none of us wants this to be merged as is?

rela589n:
Yea, I've noticed that. Though, it must've been implemented this way from the beginning.

greg0ire:
This is going in circles, let's stop the conversation here if you don't mind.

rela589n:
@greg0ire , actually I do mind

rela589n:
I've clearly stated my vision, and this's no reason to close the PR

rela589n:
@greg0ire , I really believe that this should've been implemented this way

greg0ire:
Well what's the point of reopening it if no one is going to merge it? I certainly am not…

> I really believe that this should've been implemented this way

I already got that, and already told you I got that, and I disagree, and it's not just me, so no need to reiterate, really.

rela589n:
Please, let me explain myself.
I'm reaching for a better quality code software.
I stumbled upon the problem that tests were included in non-dev environments.
Evidently, I went and searched how it migth be customized, and finally I found that it was almost there. Almost what I needed. Quite nearly, almost totally, but not.
And if there was a setter for a regex rather than setter for an extension, it would be it.

greg0ire:
I understood that as well. What _you_ don't seem to understand is that opening this can of worms opens up this repository to a lot of support requests from users who won't know how to use this properly (because let's face it, it's hard to use). In other words, it changes the public API of this library from straightforward to quite complex. What @derrabus pointed to is another way to achieve the same goal but would be user friendly, but you don't want to implement it. That's fully OK, but don't try to get us to merge this then.

rela589n:
> can of worms

This is not a can of worms. It's no good to call any regex solution bad because some might not understand regex. You might mean bugs in the code due to regex. This feature is going to be as one time action taken during project setup. If anybody needs this, he'll do it, if not - he'll not be bothered with it. This is not something that people will need to use and know much about, as they won't be thinking about it, nor using it all the time.

> a lot of support requests

Have there been support requests about regex?

> changes the public API of this library from straightforward to quite complex

It doesn't change API of the library from straightforward to complex. There isn't any API (no straightforward, no other) changed. Code opens for usage that, what already was there.

> another way to achieve the same goal
> you don't want to implement 

Yes, I do not want to implement other solution, as the implementation plan is obscure, and it requires changes in other places, and possibly other libraries, it anticipates completely different approach to the one that already exists, and it's not clear how this is supposed to work with the given paths which are already collected, and how this all is eventually expected to be  integrated with symfony anyway.

It might look interesting to have the ability to use symfony finder, but I do not find it clear, and neither reasonable or accomplishable due to the current way colocated mapping driver works.

greg0ire:
>  There isn't any API (no straightforward, no other) changed

Well the old API is deprecated, so it will be removed in the next major, so eventually, it's a change even if right now, it just looks like an addition.

> Yes, I do not want to implement other solution, as the implementation plan is obscure, 

Then clarify it. Are we the ones that have to do all the thinking to solve _your_ issue? To help you, the plan looks like this, to me:

1. In the bundles, define configuration allowing to define inclusion and exclusion patterns (as either globs or regexes), because [`symfony/finder` supports both](https://symfony.com/doc/current/components/finder.html#file-name).
2. Still in the bundles, use `symfony/finder` to build an iterable of files from the configuration nodes this.
3. Pass the iterable of files to the library.

Maybe you can get more details/validation from @derrabus since it was his idea.

rela589n:
> Well the old API is deprecated

Well, this isn't used anyway, so is looks like "straightforward useless API is changed into _quite complex_ useful API" :smile: , though I do not find it complex at all.

> Then clarify it. Are we the ones that have to do all the thinking to solve your issue? 

That's your proposed solution, and it might be fine, but if it's obscure and I don't get a point, it's just not reasonable to think that "a lot of changes" will be any better than "small regex", as far as end result remains the same

greg0ire:
> though I do not find it complex at all

The new API has an argument that is a regex, which embeds a big complexity because it's basically a small state machine, and when you look at usage examples, you see that you have to know about advanced regex concepts to make it useful. Compare that to an API that takes an iterable of filepaths.

> it's just not reasonable to think that "a lot of changes" will be any better than "small regex", as far as end result remains the same

The end result might be the same but the usage to reach that end result definitely isn't, as outlined in https://github.com/doctrine/persistence/pull/417#issuecomment-2823614201

rela589n:
I agree that the example might be complex.

Consider this instead: `(?<!Fixture)\.php$` (https://regex101.com/r/YLJJ7G/1)

Basically this is the same, and excludes everything that ends with `Fixture.php`

rela589n:
@greg0ire , Sorry for providing the complex example. I agree that it's nuts and crazy.

Given that the example is changed, can we proceed with this?

So, `^(?!.*(Test|Fixture|Command)\.php$).*\.php$` is replaced with `(?<!Test|Fixture|Command)\.php$`

rela589n:
@greg0ire , `(?<!Test|Fixture|Command)\.php$` looks simple enough, isn't it?

greg0ire:
I would say it still looks a bit nuts and crazy.

rela589n:
It's not.

rela589n:
It's not nuts, nor crazy. It's the simplest thing that could be.

greg0ire:
It's probably the simplest regex that could be for this case, I'll concede that.

rela589n:
To tell the truth, I'm glad to see it. So let's reopen it

rela589n:
@greg0ire , hi! So let's reopen it

greg0ire:
Hi! I think you misunderstood, it's probably the simplest regex, and that's not good enough, sorry.

rela589n:
@greg0ire , it's simple enough so that anybody can understand it

rela589n:
@greg0ire , how can we move on?

greg0ire:
By implementing the solution proposed by @derrabus

rela589n:
@derrabus , can we elaborate on this? What is the interface that you propose?

rela589n:
@derrabus , @greg0ire , just a reminder that I would want to move on.

rela589n:
Hi, @derrabus !

I guess you've missed the message. Could you please reply

rela589n:
@greg0ire , @derrabus , can you please respond?

PR #421, opened 2025-05-21
Fix RuntimeReflectionProperty not working with static properties

### LINKED ISSUE #420: Static properties fail to be accessed through RuntimeReflectionProperty
### Bug Report

| Version                                     | 3.3.3
| Previous Version if the bug is a regression | 2.5.7

#### Summary

After upgrading 2.5.7 → 3.3.3 as a dependency of the Doctrine ODM we are seeing crashes:

```
Cannot assign null to property RH\OrganisationBundle\Model\Organisation::$type of type string
```

The catch here is that the `$type` property is `static`. But it has been like this for years and years in our codebase and worked OK.

This happens when running the `UnitOfWork->merge()` method. I have tracked this down to the `RuntimeReflectionProperty` behaving differently from the native `ReflectionProperty` for static properties:

```php
<?php

require 'vendor/autoload.php';

class Foo
{
    public static $staticProp = 'staticPropValue';
}

$fooInstance = new Foo();

$propNative = new ReflectionProperty(Foo::class, 'staticProp');
$propDoctrine = new \Doctrine\Persistence\Reflection\RuntimeReflectionProperty(Foo::class, 'staticProp');

var_dump(
    $propNative->getValue(),
    $propNative->getValue($fooInstance),
    $propDoctrine->getValue(),
    $propDoctrine->getValue($fooInstance)
);

/*
 * string(15) "staticPropValue"
 * string(15) "staticPropValue"
 * string(15) "staticPropValue"
 * NULL
 */
```

I believe this has been introduced in https://github.com/doctrine/persistence/commit/45c484fc8d70bf5aa4b4986a4f95210adec55878.

#### Current behavior

`RuntimeReflectionProperty` returns `null` for static properties.

#### Expected behavior

`RuntimeReflectionProperty` should return actual values for static properties.

#### How to reproduce

See linked PR.

melkamar:
Fixes #420.

My assumption here is that using the native `ReflectionProperty` for static properties won't cause the lazy proxy classes to get instantiated. The tests seem to confirm that.

### DIFF
--- a/src/Persistence/Reflection/RuntimeReflectionProperty.php
+++ b/src/Persistence/Reflection/RuntimeReflectionProperty.php
@@ -38,7 +38,7 @@ class RuntimeReflectionProperty extends ReflectionProperty
     #[ReturnTypeWillChange]
     public function getValue($object = null)
     {
-        if ($object === null) {
+        if ($object === null || $this->isStatic()) {
             return parent::getValue($object);
         }
 
@@ -56,7 +56,7 @@ class RuntimeReflectionProperty extends ReflectionProperty
     #[ReturnTypeWillChange]
     public function setValue($object, $value = null)
     {
-        if (! ($object instanceof Proxy && ! $object->__isInitialized())) {
+        if ((! ($object instanceof Proxy && ! $object->__isInitialized())) || $this->isStatic()) {
             parent::setValue($object, $value);
 
             return;
--- a/tests/Persistence/RuntimeReflectionPropertyTest.php
+++ b/tests/Persistence/RuntimeReflectionPropertyTest.php
@@ -27,6 +27,7 @@ class RuntimeReflectionPropertyTest extends TestCase
     /**
      * @testWith ["test", "testValue"]
      *           ["privateTest", "privateTestValue"]
+     *           ["staticTest", "staticTestValue"]
      */
     public function testGetSetValue(string $name, string $value): void
     {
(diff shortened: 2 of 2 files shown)


src/Persistence/Reflection/RuntimeReflectionProperty.php
```diff
@@ -56,7 +56,7 @@ public function getValue($object = null)
     #[ReturnTypeWillChange]
     public function setValue($object, $value = null)
     {
-        if (! ($object instanceof Proxy && ! $object->__isInitialized())) {
+        if ((! ($object instanceof Proxy && ! $object->__isInitialized())) || $this->isStatic()) {
```
GromNaN, line 59: The extra parenthesis are not necessary. You can switch the operands to remove any ambiguity.
```suggestion
        if ($this->isStatic() || ! ($object instanceof Proxy && ! $object->__isInitialized())) {
```

GromNaN (commented):
I don't think this was a use case that had been considered, given that this class is designed to access the properties of managed objects. In fact, I don't think the latest versions of Doctrine ORM use it. What is your use case?

GromNaN:
I've seen your comment https://github.com/doctrine/persistence/issues/420#issuecomment-2911960676, I think PR can be closed? Could you open an issue on [doctrine/mongodb-odm](https://github.com/doctrine/mongodb-odm) ?

PR #422, opened 2025-05-27
fix: Add parent constructor call in EnumReflectionProperty

### LINKED ISSUE #423: EnumReflectionProperty throws Fatal Error
### Bug Report

| Version                                     | 4.0.0

Versions < 4.0.0 are more than likely also impacted by this bug.

#### Summary

Hello, I am the developer of `DoctrineDiagramBundle`. With this bundle I can generate class diagrams from my Doctrine entities.  
The problem is that when an entity has a Enum type property my bundle fails and throws a **Fatal error**.

After some hours of research I identified the problem: some methods in `\Doctrine\Persistence\Reflection\EnumReflectionProperty`  don't work properly and generate a fatal error.
More precisely, any inherited method in `EnumReflectionProperty` will generate an error.

> Internal error: Failed to retrieve the reflection object

Image

#### Current behavior

Methods defined in `EnumReflectionProperty` work as expected, for example:

* `\Doctrine\Persistence\Reflection\EnumReflectionProperty::getName` **Ok**
* `\Doctrine\Persistence\Reflection\EnumReflectionProperty::isPrivate` **Ok**

But methods inherited from parent will fail, for example:

* `\Doctrine\Persistence\Reflection\EnumReflectionProperty::hasDefaultValue` **Fatal error**
* `\Doctrine\Persistence\Reflection\EnumReflectionProperty::isPublic` **Fatal error**

#### Expected behavior

All methods from `EnumReflectionProperty` must work properly without any error.

#### How to reproduce

To reproduce the issue simply call any inherited method in `EnumReflectionProperty`.

I reproduced the issue installing and using DoctrineDiagramBundle: 
* Install [jawira/doctrine-diagram-bundle](https://packagist.org/packages/jawira/doctrine-diagram-bundle).
* Create an entity with an Enum type property, this property must also be a Doctrine property. This is described in the original issue:  
  https://github.com/jawira/entity-draw/issues/17.
* Execute `bin/console doctrine:diagram:class` and voila, you get **Fatal error**.

#### Solution

As per documentation, when a Reflection class is extended and constructor overwritten, then **the parent's constructor must also be called**.

Image

Source: <https://www.php.net/manual/en/reflection.extending.php>

**I have created a PR that fixes this issue:** https://github.com/doctrine/persistence/pull/422

jawira:
Fixes https://github.com/doctrine/persistence/issues/423

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -6,16 +6,6 @@ parameters:
 			path: src/Persistence/Mapping/Driver/StaticPHPDriver.php
 
 		-
-			message: "#^Doctrine\\\\Persistence\\\\Reflection\\\\EnumReflectionProperty\\:\\:__construct\\(\\) does not call parent constructor from ReflectionProperty\\.$#"
-			count: 1
-			path: src/Persistence/Reflection/EnumReflectionProperty.php
-
-		-
-			message: "#^Method Doctrine\\\\Persistence\\\\Reflection\\\\EnumReflectionProperty\\:\\:getDeclaringClass\\(\\) return type with generic class ReflectionClass does not specify its types\\: T$#"
-			count: 1
-			path: src/Persistence/Reflection/EnumReflectionProperty.php
-
-		-
 			message: "#^Method Doctrine\\\\Persistence\\\\Reflection\\\\EnumReflectionProperty\\:\\:toEnum\\(\\) should return array\\<BackedEnum\\>\\|BackedEnum but returns array\\<BackedEnum\\|int\\|string\\>\\.$#"
 			count: 1
 			path: src/Persistence/Reflection/EnumReflectionProperty.php
--- a/src/Persistence/Reflection/EnumReflectionProperty.php
+++ b/src/Persistence/Reflection/EnumReflectionProperty.php
@@ -5,9 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Persistence\Reflection;
 
 use BackedEnum;
-use ReflectionClass;
 use ReflectionProperty;
-use ReflectionType;
 
 use function array_map;
 use function is_array;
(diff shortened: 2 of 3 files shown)


SenseException (dismissed):

jawira:
Hi! is there anything I could do to help get this PR merged.  
All checks have passed.  
Thanks! :)

greg0ire:
> Once EnumReflectionProperty::__contruct calls its parent constructor (as in PR https://github.com/doctrine/persistence/pull/422) all methods like the following one can be safely removed (there's six of them):

Why don't you do it rightaway? Also, will there still be a point in keeping `private readonly ReflectionProperty $originalReflectionProperty` as a property if we do that?

I think this clashes a bit with the original design made in #248 by @IonBazan

jawira:
Ah you are right, I spoke too quickly. I cannot remove constructor since method's signature is not the same as parent class.  
However, I removed all "proxy methods" because they are not needed anymore :)

jawira:
Hello I have implemented all requested changes, could you accept this MR please ?

greg0ire:
It seems there are CI jobs failing. Please take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools) for more on how to handle those.

SenseException (dismissed):

greg0ire:
@IonBazan please review :pray:

IonBazan:
Thanks @jawira - LGTM. However, my changes were actually based on @beberlei https://github.com/doctrine/orm/pull/9304 so it's best to confirm once more with him.

Also good to add tests covering the changes you've made, reproducing the issue you reported.

IonBazan (approved):

jawira:
Hello @IonBazan, in **doctrine/orm**, `ReflectionEnumProperty` **DO** call the parent class.  
This change was made in https://github.com/doctrine/orm/pull/9381  
The code has evolved since, so I have updated the current PR to match the following:

https://github.com/doctrine/orm/blob/2c01dac17367c203de6fab0331c82eca0d639cdc/src/Mapping/ReflectionEnumProperty.php#L14-L27

I have also added tests to cover errors I had.

beberlei:
I am not sure anymore why I did it this way originally, but if this works here then great. However be mindful that for Doctrine ORM 3.4+ above we changed the abstraction completly and this code here becomes obsolete.

greg0ire (approved):

greg0ire:
Thanks @jawira !

PR #430, opened 2025-08-05
Fix: prevent `$sourceFilePathNames` from shadowing retroactively adde…

rela589n:
This PR fixes a possible bug introduced after https://github.com/doctrine/persistence/pull/429/

If the client code had called `addPaths()` after the instance creation, it would've been ignored (without any warning or so), and paths would not have been used. This commit fixes that. Retroactively added paths will be taken into account as well as `$sourceFilePathNames`.

### DIFF
--- a/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
+++ b/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
@@ -19,7 +19,6 @@ use function array_merge;
 use function array_unique;
 use function assert;
 use function get_declared_classes;
-use function in_array;
 use function is_dir;
 use function preg_match;
 use function preg_quote;
@@ -165,7 +164,8 @@ trait ColocatedMappingDriver
             $this->sourceFilePathNames ?? [],
             $this->pathNameIterator($filesIterator),
         );
-        $includedFiles       = [];
+        /** @var array<string,true> $includedFiles */
+        $includedFiles = [];
 
         foreach ($sourceFilePathNames as $sourceFile) {
             if (preg_match('(^phar:)i', $sourceFile) === 0) {
@@ -186,7 +186,7 @@ trait ColocatedMappingDriver
 
             require_once $sourceFile;
 
-            $includedFiles[] = $sourceFile;
+            $includedFiles[$sourceFile] = true;
         }
 
         $classes  = [];
@@ -197,7 +197,7 @@ trait ColocatedMappingDriver
 
             $sourceFile = $rc->getFileName();
 
-            if (! in_array($sourceFile, $includedFiles, true) || $this->isTransient($className)) {
+            if (! isset($includedFiles[$sourceFile]) || $this->isTransient($className)) {
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


greg0ire:
Please make sure the subject of your commit message are no longer than 50 chars, and please wrap the body of your commit messages at 72 chars.

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -217,4 +220,24 @@ private function pathNameIterator(iterable $filesIterator): Generator
...
+     * @return Generator<TKey, T>
+     *
+     * @template TKey
+     * @template T
+     */
+    private function mergeIterables(iterable $iterable1, iterable $iterable2): Generator
+    {
+        foreach ($iterable1 as $key => $item) {
+            yield $key => $item;
+        }
```
greg0ire: I think this can be rewritten as

```suggestion
        yield from $iterable1;
```

Same below.

rela589n:
> Please make sure the subject of your commit message are no longer than 50 chars, and please wrap the body of your commit messages at 72 chars.

Could you please check it right now? I've shortened the commit messages, but the last commit message is not quite up to par. If needed, I can try to shorten it even more, but it would lose the context then.

greg0ire:
Why do you say you would lose the context? Can't you move it from the commit message subject to the commit message body? You did read that I asked to wrap the commit message body at 72 chars? Wrap, not truncate.

rela589n:
> Wrap, not truncate.

Sorry, didn't get your point. 
What do you mean by "wrap"? 
I thought I've wrapped exactly how you'd asked me to.

greg0ire:
Wrapping just means adding a newline/carriage return every 72 chars.

greg0ire:
Now that I am on desktop, it does look wrapped just fine 👍

So why do you say you would lose the context? Can't you move it from the commit message subject to the commit message body?

rela589n:
> So why do you say you would lose the context? Can't you move it from the commit message subject to the commit message body?

Done

greg0ire (dismissed):

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -207,14 +210,17 @@ public function getAllClassNames(): array
...
+     * @param iterable<TKey, T> $iterable2
      *
-     * @return Generator<int,string>
+     * @return Generator<TKey, T>
+     *
+     * @template TKey
+     * @template T
      */
-    private function pathNameIterator(iterable $filesIterator): Generator
+    private function mergeIterables(iterable $iterable1, iterable $iterable2): Generator
```
SenseException, line 221: This is rather appending `$iterable2` to `$iterable1` than merging.

SenseException (approved):

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
rela589n: Do you propose naming it as `concatIterables()`? 

"Appending" sounds with the tone of modifying the iterable, while it's actually combining them into a single iterable, doing no modification to the original.

Dictionary-wise, "merge" looks good as it means "combine or cause to combine to form a single entity."

greg0ire:
Closing in favor of a better incoming PR

PR #432, opened 2025-08-08
Add support for file path iterables in `ColocatedMappingDriver`

rela589n:
### Description

This PR follows from https://github.com/doctrine/persistence/pull/429#issuecomment-3167293606

- **Feature:** Added support for `ColocatedMappingDriver` to accept `iterable` of file paths. This enhances flexibility as clients can now provide specific iterators of required files, rather than an array of complete directories.
- **Refactor:** Introduced `DirectoryFilesIterator` and `FilePathNameIterator` for easier integration with iterable file paths. This benefits libraries by adapting `SplFileInfo` objects.
- **Optimization:** Improved file lookup performance by replacing `in_array()` within a loop (O(N * M)) with `isset()` (O(N)), providing a significant performance enhancement.

### DIFF
--- a/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
+++ b/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
@@ -7,7 +7,6 @@ namespace Doctrine\Persistence\Mapping\Driver;
 use AppendIterator;
 use Doctrine\Persistence\Mapping\MappingException;
 use FilesystemIterator;
-use Generator;
 use Iterator;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
@@ -156,7 +155,8 @@ trait ColocatedMappingDriver
             $filesIterator->append($iterator);
         }
 
-        $sourceFilePathNames = $this->pathNameIterator($filesIterator);
+        /** @var iterable<string> $sourceFilePathNames */
+        $sourceFilePathNames = new FilePathNameIterator($filesIterator);
         /** @var array<string,true> $includedFiles */
         $includedFiles = [];
 
@@ -200,17 +200,5 @@ trait ColocatedMappingDriver
         $this->classNames = $classes;
 
         return $classes;
-    }
-
-    /**
-     * @param iterable<SplFileInfo> $filesIterator
-     *
-     * @return Generator<int,string>
-     */
-    private function pathNameIterator(iterable $filesIterator): Generator
-    {
-        foreach ($filesIterator as $file) {
-            yield $file->getPathname();
-        }
     }
 }
--- a/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
+++ b/src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
@@ -21,8 +21,11 @@ use function str_replace;
  */
 trait ColocatedMappingDriver
 {
+    /** @var iterable<array-key,string> */
+    private iterable $filePaths;
+
     /**
-     * The paths where to look for mapping files.
+     * The directory paths where to look for mapping files.
      *
      * @var array<int, string>
      */
@@ -39,7 +42,7 @@ trait ColocatedMappingDriver
     protected string $fileExtension = '.php';
 
     /**
-     * Cache for getAllClassNames().
+     * Cache for {@see getAllClassNames()}.
      *
      * @var array<int, string>|null
      * @phpstan-var list<class-string>|null
@@ -67,7 +70,7 @@ trait ColocatedMappingDriver
     }
 
     /**
-     * Append exclude lookup paths to metadata driver.
+     * Append exclude lookup paths to a metadata driver.
      *
      * @param string[] $paths
      */
@@ -120,13 +123,17 @@ trait ColocatedMappingDriver
             return $this->classNames;
         }
 
-        if ($this->paths === []) {
+        if ($this->paths === [] && ! isset($this->filePaths)) {
             throw MappingException::pathRequiredForDriver(static::class);
         }
 
         $dirFilesIterator = new DirectoryFilesIterator($this->paths, $this->fileExtension);
+
         /** @var iterable<string> $filePathsIterator */
-        $filePathsIterator = new FilePathNameIterator($dirFilesIterator);
+        $filePathsIterator = $this->concatIterables(
+            $this->filePaths ?? [],
+            new FilePathNameIterator($dirFilesIterator),
+        );
 
         /** @var array<string,true> $includedFiles */
         $includedFiles = [];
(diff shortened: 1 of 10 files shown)


src/Persistence/Mapping/Driver/DirectoryFilesIterator.php
greg0ire, line 1: Let's mark this file as `@internal`, since it's going to die in the future?

src/Persistence/Mapping/Driver/FilePathNameIterator.php
greg0ire, line 1: For now I see no reason to expose this file as well => internal

tests/Persistence/Mapping/ColocatedMappingDriverTest.php
```diff
@@ -48,46 +50,86 @@ public function testAddGetExcludePaths(): void
...
     }
 }
 
 final class MyDriver implements MappingDriver
 {
     use ColocatedMappingDriver;
 
-    /** @param non-empty-list<string> $paths One or multiple paths where mapping classes can be found. */
-    public function __construct(array $paths)
+    /** @param non-empty-list<string>|Traversable<string> $paths One or multiple paths where mapping classes can be found. */
```
greg0ire:
```suggestion
    /** @param non-empty-list<string>|iterable<string> $paths One or multiple paths where mapping classes can be found. */
```

If it's an array, then it should be OK to take a look at `$paths[0]` and determine if it is a file or a directory.
If it's not an array, you can assume it goes to `$filePaths`.

tests/Persistence/Mapping/ColocatedMappingDriverTest.php
```diff
@@ -48,46 +50,86 @@ public function testAddGetExcludePaths(): void
...
+    public function __construct(iterable $paths)
     {
-        $this->addPaths($paths);
+        $isFilePaths = $paths instanceof Traversable;
+
+        if (! $isFilePaths) {
+            $this->paths = $paths;
+        } else {
+            $this->filePaths = $paths;
+        }
```
greg0ire:
```suggestion
    if ($paths instanceof Traversable || is_file($paths[0]) {
        $this->filePaths = $paths;
        
        return;
    }
    
    $this->paths = $paths;
```

tests/Persistence/Mapping/ColocatedMappingDriverTest.php
```diff
@@ -48,46 +50,86 @@ public function testAddGetExcludePaths(): void
...
+    private function createDirectoryPathDriver(string $dirPath): MyDriver
     {
-        return new MyDriver([$path]);
+        return new MyDriver([$dirPath]);
+    }
+
+    /** @param list<string> $filePaths */
+    private function createFilePathsDriver(array $filePaths): MyDriver
+    {
+        return new MyDriver(new ArrayIterator($filePaths));
```
greg0ire: No, this is too ugly, sorry. You don't want to document something similar in ORM and ODM. The fact that a `symfony/finder` is an `Iterator` is cool, but it shouldn't become a requirement. The ORM and ODM must stay usable without the Symfony bundles.

```suggestion
        return new MyDriver($filePaths);
```

src/Persistence/Mapping/Driver/FilePathNameIterator.php
rela589n: Do you think it's worth making it internal? It would be necessary to implement this same class again where we would like to pass Symfony Finder, because Finder doesn't represent `iterable<string>`, but instead it represents `iterable<SplFileInfo>`. I expected to use this class as an adapter

tests/Persistence/Mapping/ColocatedMappingDriverTest.php
rela589n: Agreeing with you. Your provided solution is neat, so I'll stick with it.

src/Persistence/Mapping/Driver/DirectoryFilesIterator.php
rela589n: Do you think it's a good idea to rely solely on Symfony Finder ?

src/Persistence/Mapping/Driver/FilePathNameIterator.php
greg0ire: Ah if you expect to use this class from the outside, then ignore my comment.

src/Persistence/Mapping/Driver/DirectoryFilesIterator.php
greg0ire: :thinking: you mean we should expose this as the default implementation? It might make sense, yeah. Let's keep it public then.

greg0ire (dismissed):

src/Persistence/Mapping/Driver/DirectoryFilesIterator.php
GromNaN, line 1: For ease of maintenance, could you add unit tests to the new classes  `FilePathNameIterator` and `DirectoryFilesIterator`?

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -132,35 +123,22 @@ public function getAllClassNames(): array
...
-                RegexIterator::MATCH,
-            );
-
-            $filesIterator->append($iterator);
-        }
+        /** @var iterable<string> $filePathsIterator */
+        $filePathsIterator = $this->concatIterables(
+            $this->filePaths ?? [],
+            new FilePathNameIterator($dirFilesIterator),
+        );
```
GromNaN, line 136: The `AppendIterator` would be perfect to replace `concatIterables` and avoid this private method to the class that uses the trait.
```suggestion
        $filePathsIterator = new \AppendIterator();
        $filePathsIterator->append(new ArrayIterator($this->filePaths ?? []));
        $filePathsIterator->append(new FilePathNameIterator($dirFilesIterator));
```

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -203,14 +181,17 @@ public function getAllClassNames(): array
...
+     * @param iterable<TKey, T> $iterable1
+     * @param iterable<TKey, T> $iterable2
+     *
+     * @return iterable<TKey, T>
      *
-     * @return Generator<int,string>
+     * @template TKey
+     * @template T
      */
-    private function pathNameIterator(iterable $filesIterator): Generator
```
GromNaN, line 210: Technically, removing a private method from a trait is a breaking change. It could be used by the class that uses it in an other package. But this code has not been released yet.

That's why I thing it's better not introducing `concatIterables` in this trait.

GromNaN (commented):
I really like this progressive approach.

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
rela589n: I also considered doing it this way, but the problem is that `ArrayIterator` doesn't accept a generic `iterable` (nor `Traversable`). Although it works well with `array|ArrayObject`, that's the only input type it supports. It won't work with `$filePaths` being some `Traversable` instead.

GromNaN:
Replaced by #433. Thank you @rela589n for exploring this topic.

PR #433, opened 2025-08-12
Introduce `ClassLocator` to find class names for attribute drivers

GromNaN:
Alternative approach to https://github.com/doctrine/persistence/pull/432 to decouple the class name discovery from the read of metadata in the attribute (and annotation) drivers.

Other drivers use a [`FileLocator`](https://github.com/doctrine/persistence/blob/4.1.x/src/Persistence/Mapping/Driver/FileLocator.php) to provide the list of files to read.

For the attribute/annotation/static-php drivers, the driver class itself does the directory scanning, while its main purpose is to load the metadata of each class.

In this PR I introduce a `ClassLocator` interface with 2 implementations:
- `ClassNames` is a simple list of classes names
- `FileClassLocator` scans a list of files, it can also be constructed from a list of directories to support the current usage. `FileClassLocator::createFromDirectories($dirPaths, $fileExtension, $excludedDirs)`.

My goal is to make it possible in Symfony to tag any class in the project with the `#[Entity]` attribute and get it identified by the EntityManager. [Symfony DI resource discovery feature](https://symfony.com/blog/new-in-symfony-7-3-dependency-injection-improvements#resource-tags) will be leveraged so that we don't need to scan the directories in Doctrine (which is done at runtime, and may be cached).


GromNaN:
@rela589n This PR goes a step further than what you've proposed, by exposing a list of class names instead of scanning file.

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * This method can be used with a Symfony Finder or any other iterable
+     *
+     * @param iterable<SplFileInfo> $files
+     */
+    public static function createFromSplFiles(iterable $files): self
+    {
+        return new self((static function ($files) {
+            foreach ($files as $file) {
+                // @phpstan-ignore-next-line
```
greg0ire: Can you please use instead `phpstan-ignore issue.identifier (reason for the issue)`?

src/Persistence/Mapping/Driver/ClassLocator.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Doctrine\Persistence\Mapping\Driver;
+
+/**
+ * ClassLocator is an interface for classes that can provide a list of class names.
+ */
+interface ClassLocator
+{
+    /** @return class-string[] */
+    public function getClassNames(): array;
```
greg0ire, line 13: What justifies providing an array and not just an iterable?

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -123,75 +116,24 @@ public function getAllClassNames(): array
...
-     *
-     * @template TKey
-     * @template T
-     */
-    private function concatIterables(iterable $iterable1, iterable $iterable2): iterable
-    {
-        yield from $iterable1;
-        yield from $iterable2;
+        return $this->classNames = array_filter(
+            array_unique($classLocator->getClassNames()),
```
greg0ire: In what real world scenario is this useful?

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,156 @@
...
+                assert($file instanceof SplFileInfo, new InvalidArgumentException(sprintf('Expected an iterable of SplFileInfo, got %s', get_debug_type($file))));
```
greg0ire: TIL you can use a `Throwable` here :bulb:

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Creates a FileClassLocator from an array of directories.
+     *
+     * @param list<string> $directories
+     * @param string       $fileExtension       The file extension to look for (default is '.php').
+     * @param list<string> $excludedDirectories Directories to exclude from the search.
+     *
+     * @throws MappingException if any of the directories are not valid.
+     */
+    public static function createFromDirectories(array $directories, string $fileExtension = '.php', array $excludedDirectories = []): self
```
greg0ire: Please use linebreaks.

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,156 @@
...
+            );
+        }
+
+        return self::createFromSplFiles($filesIterator);
+    }
+
+    /**
+     * Creates a FileClassLocator from an iterable of SplFileInfo objects.
+     *
+     * This method can be used with a Symfony Finder or any other iterable
```
greg0ire: Should we require `symfony/finder` as a dev dependency and prove this?

src/Persistence/Mapping/Driver/DirectoryFilesIterator.php
greg0ire, line 1: Where is this class supposed to be used?

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: I agree that using the precise identifier is better. I've just learned that I need to use the verbose output in order to get the identifier in the output `vendor/bin/phpstan -v`

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: Good idea, dev-dependency added.

src/Persistence/Mapping/Driver/ClassLocator.php
GromNaN: Because `MappingDriver::getAllClassNames()` returns an `array`. That's much simpler to work with an array than an iterator.
https://github.com/doctrine/persistence/blob/275afd6b2cbb1c8249d2c3af3d19054fc52f5dfd/src/Persistence/Mapping/Driver/MappingDriver.php#L30

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
GromNaN: If both the directory paths and the class locator are set, it could contain duplicates.
Imagine adding support for DIC discovery and keeping the legacy directory scanning, the same files will be listed twice.
I've moved this call to the array merge.

src/Persistence/Mapping/Driver/DirectoryFilesIterator.php
GromNaN: Leftover, removed.

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
greg0ire: Ok, but that would be the result of some kind of misconfiguration, right? Should we throw in case there are duplicates instead?

GromNaN: In my opinion, it's not a problem if a class is found twice, since class names can be deduplicated. And it allows for more flexibility in configuration.

greg0ire: Ok, let's keep it this way then 👍

greg0ire (dismissed):

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -123,75 +116,24 @@ public function getAllClassNames(): array
...
+                    ...$this->classLocator->getClassNames(),
+                ]);
             }
-
-            $classes[] = $className;
+        } elseif (isset($this->classLocator)) {
+            $classNames = $this->classLocator->getClassNames();
+        } else {
+            throw MappingException::pathRequiredForDriver(static::class);
         }
```
rela589n: Generally, early returns and early throws read better
If you wish, you could simplify it:

```diff
             return $this->classNames;
         }

-        if ($this->paths !== []) {
-            $classNames = FileClassLocator::createFromDirectories($this->paths, $this->excludePaths, $this->fileExtension)->getClassNames();
-            if (isset($this->classLocator)) {
-                $classNames = array_unique([
-                    ...$classNames,
-                    ...$this->classLocator->getClassNames(),
-                ]);
-            }
-        } elseif (isset($this->classLocator)) {
-            $classNames = $this->classLocator->getClassNames();
-        } else {
-            throw MappingException::pathRequiredForDriver(static::class);

+        if ($this->paths === [] && ! isset($this->classLocator)) {
+            throw MappingException::pathRequiredForDriver(static::class);
+        }
+
+        $classNames = FileClassLocator::createFromDirectories($this->paths, $this->excludePaths, $this->fileExtension)->getClassNames();
+
+        if (isset($this->classLocator)) {
+            $classNames = array_unique([
+                ...$classNames,
+                ...$this->classLocator->getClassNames(),
+            ]);
         }

         return $this->classNames = array_filter(
```

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,174 @@
...
+            );
+
+            $filesIterator = new CallbackFilterIterator(
+                $filesIterator,
+                static function (SplFileInfo $file) use ($excludedDirectories): bool {
+                    if (str_starts_with($file->getPath(), 'phar:')) {
+                        $sourceFile = $file->getPathname();
+                    } else {
+                        $sourceFile = $file->getRealPath();
+                    }
```
rela589n: For some reason, the original code had `(^phar:)i` check as case-insensitive. Are we sure that `phar:` isn't likely to be caps?

By the way, can't we use `getRealPath()` all the time?

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
```diff
@@ -123,75 +116,24 @@ public function getAllClassNames(): array
...
-     * @template T
-     */
-    private function concatIterables(iterable $iterable1, iterable $iterable2): iterable
-    {
-        yield from $iterable1;
-        yield from $iterable2;
+        return $this->classNames = array_filter(
+            $classNames,
+            fn (string $className): bool => ! $this->isTransient($className),
+        );
```
rela589n: `array_filter` is real good :+1: 

But make sure that it returns the expected type of array, since it doesn't reset the keys

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,174 @@
...
+
+            $includedFiles[$fileName] = true;
+            require_once $fileName;
+        }
+
+        $classes = [];
+        foreach (get_declared_classes() as $className) {
+            $fileName = (new ReflectionClass($className))->getFileName();
+
+            if ($fileName === false || ! array_key_exists($fileName, $includedFiles)) {
```
rela589n, line 71: Using `! isset($includedFiles[$fileName])` would be faster than `array_key_exists`

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,174 @@
...
+{
+    /**
+     * @param iterable<string> $fileNames An iterable of file names to include.
+     *
+     * @throws MappingException if any of the files do not exist.
+     */
+    public function __construct(
+        private iterable $fileNames,
+    ) {
+    }
```
rela589n: Constructor doesn't throw anything, does it?
Probably `MappingException` should've been described on `getClassNames()`

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        if ($excludedDirectories !== []) {
+            // Get
+            $excludedDirectories = array_map(
+                // @phpstan-ignore ternary.shortNotAllowed
+                static fn (string $dir): string => realpath($dir) ?: $dir,
+                $excludedDirectories,
+            );
+
```
rela589n: There was also originally a logic that replaces backslashes with forward slashes

```php
$exclude = str_replace('\\', '/', $realExcludePath);
$current = str_replace('\\', '/', $sourceFile);
```

Do you think it's all right to do w/o it?

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,174 @@
...
+    {
+        $includedFiles = [];
+
+        foreach ($this->fileNames as $fileName) {
+            if (! is_file($fileName)) {
+                throw MappingException::fileDoesNotExist($fileName);
+            }
+
+            // realpath() can return false if the file is in a phar archive
+            // @phpstan-ignore ternary.shortNotAllowed
```
rela589n: Using `@phpstan-ignore` isn't working fine with lowest dependencies (`composer update --prefer-lowest`):

```
------ ----------------------------------------------------------------------------------------------------------------- 
 Line   src/Persistence/Mapping/Driver/FileClassLocator.php                                                              
------ ----------------------------------------------------------------------------------------------------------------- 
 61     No error with identifier ternary.shortNotAllowed is reported on line 61.                                         
 61     Short ternary operator is not allowed. Use null coalesce operator if applicable or consider using long ternary.  
```

image

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,174 @@
...
+            $fileName = realpath($fileName) ?: $fileName;
+            if (isset($includedFiles[$fileName])) {
+                continue;
+            }
+
```
rela589n: Do you think it's worth checking for an already included file, since `require_once` already does it?

tests/Persistence/Mapping/ColocatedMappingDriverTest.php
```diff
@@ -112,23 +112,21 @@ private function createDirectoryPathDriver(string $dirPath): MyDriver
...
-        $isFilePaths = $paths instanceof Traversable || is_file($paths[0]);
-
-        if (! $isFilePaths) {
-            $this->paths = $paths;
+        if ($paths instanceof ClassLocator) {
+            $this->classLocator = $paths;
         } else {
-            $this->filePaths = $paths;
+            $this->paths = $paths;
         }
```
rela589n: This is far better than `Traversable` :+1:

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,174 @@
...
+     * This method can be used with a Symfony Finder or any other iterable
+     *
+     * @param iterable<SplFileInfo> $files
+     */
+    public static function createFromSplFiles(iterable $files): self
+    {
+        return new self((static function ($files) {
+            foreach ($files as $file) {
+                // @phpstan-ignore function.alreadyNarrowedType, instanceof.alwaysTrue
+                assert($file instanceof SplFileInfo, new InvalidArgumentException(sprintf('Expected an iterable of SplFileInfo, got %s', get_debug_type($file))));
```
rela589n: @GromNaN , @greg0ire , how do you think maybe it would make sense for `FileClassLocator` to accept the iterable of `SplFileInfo`? It would remove the need for `createFromSplFiles()` and simplify `getClassNames()` as `SplFileInfo` interface provides easy-to-use methods: `isFile()`, `isDir()`, `getRealPath()`

tests/Persistence/Mapping/Driver/FileClassLocatorTest.php
```diff
@@ -0,0 +1,163 @@
...
+        );
+
+        $classes = $locator->getClassNames();
+        sort($classes, SORT_STRING);
+
+        self::assertSame([
+            Entity::class,
+            EntityFixture::class,
+            'Doctrine\Tests\Persistence\Mapping\_files\colocated\Foo',
+            TestClass::class,
```
rela589n: It would look nicer if imported

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,174 @@
...
+            // realpath() can return false if the file is in a phar archive
+            // @phpstan-ignore ternary.shortNotAllowed
+            $fileName = realpath($fileName) ?: $fileName;
+            if (isset($includedFiles[$fileName])) {
+                continue;
+            }
+
+            $includedFiles[$fileName] = true;
+            require_once $fileName;
+        }
```
rela589n, line 65: Just to be sure, there's no guarantee that `$fileName` is a PHP file?

tests/Persistence/Mapping/Driver/FileClassLocatorTest.php
```diff
@@ -0,0 +1,163 @@
...
+        $phar->addFromString('Foo.mphp', '<?php namespace Doctrine\Phar; class Foo {}');
+        $phar->stopBuffering();
+
+        $locator = FileClassLocator::createFromDirectories(['phar://' . $pharFile], ['phar://' . $pharFile . '/Excluded']);
+
+        $classes = $locator->getClassNames();
+        sort($classes, SORT_STRING);
+        unlink($pharFile);
+
+        // @phpstan-ignore staticMethod.impossibleType
```
rela589n, line 147: I see that most of these errors are suppressed in `phpstan-baseline.neon`.
Maybe it would be better to keep up with it and `--generate-baseline` instead?

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: `getRealPath()` returns `false` for file paths in phar files.

rela589n: I saw you've used a short ternary operator between `getRealPath()` and `getPathname()`
Do you think it can be used here?

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: Since PHP 7.4, `array_key_exists` is optimized by the compiler. It's almost as fast as `isset`.

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: I initially thought about that, but if we want to create `FileClassLocator` from a list of file paths, we then have to create `SplFileInfo` objects for each path, which doesn't seem efficient to me.

rela589n: Do you think we'd want to do this? 
Wouldn't it be easier to create an instance of `ClassNames` with a list instead?

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
GromNaN: Good point, I've modified it to process the `ClassLocator` first and the dir paths conditionally when they are defined, as it is more resource consuming.

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
GromNaN: I added a call to `array_values` and I modified the `testGetAllClassNamesForFilePaths` too ensure there is no gap in the index numbers.

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: I removed this annotation. This exception is not made to be handled.

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: The error identifier was added in https://github.com/phpstan/phpstan-strict-rules/commit/477f53a560823cef4751429d91d62630ce265992, I update the dependency.

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: I removed this check, `require_once` already ensure we don't load a file that has already been loaded by PHP.

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: I think that's for Windows support. I added this transformation.

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: Good catch. So that we don't have to check for the `phar:` scheme. And we may support other schemes.

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: We check that it's a file, but we cannot check if it's a PHP file. That's the user responsibility to provide a list a PHP files only.

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: Good point, if we know the list of files, listing the classes names is probably easier.

tests/Persistence/Mapping/Driver/FileClassLocatorTest.php
GromNaN: The `Doctrine\Tests\Persistence\Mapping\_files\colocated\Foo` class is not detected by PHPStorm, because it's not in a `*.php` file. I added the file to phpstan.neon to have fix the detection of this class.

tests/Persistence/Mapping/Driver/FileClassLocatorTest.php
GromNaN: IMO the baseline if a tool to help setting up phpstan without having to fix all the issues. But for new code, the `@phpstan-ignore` comments are here to express that we intentionally went against a rule.

GromNaN (commented):
Thanks for your review @rela589n, I think I have addressed everything.

greg0ire (dismissed):

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -36,13 +36,9 @@
  */
 final class FileClassLocator implements ClassLocator
 {
-    /**
-     * @param iterable<string> $fileNames An iterable of file names to include.
-     *
-     * @throws MappingException if any of the files do not exist.
-     */
+    /** @param iterable<SplFileInfo> $files An iterable of file names to include. */
```
rela589n:
```suggestion
    /** @param iterable<SplFileInfo> $files An iterable of files to include. */
```

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -116,21 +113,17 @@ public static function createFromDirectories(
...
 
             $filesIterator = new CallbackFilterIterator(
                 $filesIterator,
                 static function (SplFileInfo $file) use ($excludedDirectories): bool {
-                    if (str_starts_with($file->getPath(), 'phar:')) {
-                        $sourceFile = $file->getPathname();
-                    } else {
-                        $sourceFile = $file->getRealPath();
-                    }
+                    // @phpstan-ignore ternary.shortNotAllowed
```
rela589n: As `?:` is used, anybody else who might want to do the same `?:` would be forced to add the same `@phpstan-ignore` annotation.

Maybe it would be better to ignore `ternary.shortNotAllowed` on a project-wide basis?

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -51,17 +47,18 @@ public function getClassNames(): array
...
-            if (! is_file($fileName)) {
-                throw MappingException::fileDoesNotExist($fileName);
+        foreach ($this->files as $file) {
+            // @phpstan-ignore function.alreadyNarrowedType, instanceof.alwaysTrue
+            assert($file instanceof SplFileInfo, new InvalidArgumentException(sprintf('Expected an iterable of SplFileInfo, got %s', get_debug_type($file))));
+
+            // Skip non-files
+            if (! $file->isFile()) {
+                continue;
             }
```
rela589n: Are there any situations in which `$file` could not be a file? 
Perhaps when it's not a file, it's due to some misconfiguration (`$finder->files()` returns only files)

src/Persistence/Mapping/Driver/ColocatedMappingDriver.php
rela589n: Looks good :+1:

rela589n (approved):
LGTM :+1:

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: When using a `DirectoryIterator`, you get the current directory in the results (which is a [`DirectoryIterator`](https://www.php.net/manual/en/class.directoryiterator.php) instance that extends `SplFileInfo`).

I think that's more helpful to skip non-files instead of throwing an exception.

rela589n: Ok, ignore this comment

src/Persistence/Mapping/Driver/FileClassLocator.php
GromNaN: Using `?:` is specific here because `$file->getRealPath()` returns `false`. It should have returned `null` so that `??` would have been used.

rela589n: I agree that it should've returned `null`, yet it's quite common that PHP functions return `false` instead. If you wish, you can keep it as is, but I'd rather allow `?:`

greg0ire: More background as to why this is disallowed: https://github.com/phpstan/phpstan-strict-rules/pull/92

I don't think we should allow it globally because of that.

src/Persistence/Mapping/Driver/FileClassLocator.php
```diff
@@ -0,0 +1,141 @@
...
+            // @phpstan-ignore function.alreadyNarrowedType, instanceof.alwaysTrue
+            assert($file instanceof SplFileInfo, new InvalidArgumentException(sprintf('Expected an iterable of SplFileInfo, got %s', get_debug_type($file))));
+
+            // Skip non-files
+            if (! $file->isFile()) {
+                continue;
+            }
+
+            // realpath() can return false if the file is in a phar archive
+            // @phpstan-ignore ternary.shortNotAllowed
```
greg0ire: I have no idea if this is actually true, but here is a comment showing why this should be OK.

```suggestion
            // @phpstan-ignore ternary.shortNotAllowed (false is the only falsy value getRealPath() may return)
```

GromNaN: I think the comment just above is sufficient, don't you?

greg0ire: The comment above justifies why we should check for `false`, but not why it's OK to use loose comparison.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @GromNaN !

PR #437, opened 2025-09-26
fix compatibility with PHP 8.5

xabbuh:


greg0ire (dismissed):
LGTM after reading https://www.php.net/manual/en/reflectionproperty.setaccessible.php

xabbuh:
By the way, I wonder if a job for PHP 8.4 is missing intentionally or if that’s just an oversight and should be added.

GromNaN (dismissed):

GromNaN:
> By the way, I wonder if a job for PHP 8.4 is missing intentionally or if that’s just an oversight and should be added.

Yes, I think nobody added it when PHP 8.4 was released. It was only added for 4.1 in https://github.com/doctrine/persistence/commit/5eece5b4419bcab991b5ed7d789a9146b3dd4ce4. We can add both 8.4 and 8.5?

xabbuh:
> We can add both 8.4 and 8.5?

8.4 should not be an issue. Not sure about 8.5 as I have no idea what the Doctrine policy is for adding jobs for yet to be released PHP versions.

GromNaN:
[PHP 8.5-RC1](https://www.php.net/archive/2025.php#2025-09-25-3) being released, there should be no breaking changes.

derrabus:
Let's add both.

xabbuh:
both added

GromNaN (approved):

SenseException (approved):

PR #441, opened 2025-10-07
Add `PropertyAccessor` interface

GromNaN:
Import the interface from ORM: https://github.com/doctrine/orm/blob/3.5.x/src/Mapping/PropertyAccessors/PropertyAccessor.php

### DIFF
--- a/src/Persistence/Mapping/ClassMetadataFactory.php
+++ b/src/Persistence/Mapping/ClassMetadataFactory.php
@@ -4,9 +4,12 @@ declare(strict_types=1);
 
 namespace Doctrine\Persistence\Mapping;
 
+use Doctrine\Persistence\PropertyAccessor;
+
 /**
  * Contract for a Doctrine persistence layer ClassMetadata class to implement.
  *
+ * @method PropertyAccessor createPropertyAccessor(string $className, string $propertyName)
  * @template T of ClassMetadata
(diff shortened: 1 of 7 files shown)


src/Persistence/Mapping/ClassMetadataFactory.php
```diff
@@ -4,9 +4,12 @@
 
 namespace Doctrine\Persistence\Mapping;
 
+use Doctrine\Persistence\PropertyAccessor;
+
 /**
  * Contract for a Doctrine persistence layer ClassMetadata class to implement.
  *
+ * @method PropertyAccessor createPropertyAccessor(string $className, string $propertyName)
```
GromNaN, line 12: This is a static function in ORM: https://github.com/doctrine/orm/blob/3.5.x/src/Mapping/PropertyAccessors/PropertyAccessorFactory.php

It's used in 3rd party packages like 
https://github.com/doctrine-extensions/DoctrineExtensions/blob/eb53dfcb2b592327b76ac5226fbb003d32aea37e/src/Tree/Strategy/ORM/Closure.php#L137-L151

But we would have to inject the ClassMetadataFactory into ClassMetadata...

GromNaN:
Closing as I don't find any use-case for this interface in both ORM and ODM.

PR #443, opened 2025-10-09
Upgrade to PHPStan 2.1.30

greg0ire:


composer.json
```diff
@@ -25,9 +25,9 @@
         "psr/cache": "^1.0 || ^2.0 || ^3.0"
     },
     "require-dev": {
-        "phpstan/phpstan": "1.12.7",
-        "phpstan/phpstan-phpunit": "^1",
-        "phpstan/phpstan-strict-rules": "^1.1",
+        "phpstan/phpstan": "^1 || 2.1.30",
```
SenseException, line 28: Why version 1?

SenseException (approved):

composer.json
greg0ire: Because PHP 7.2 is supported: https://github.com/doctrine/persistence/blob/4fec5bf22450ed05a8d44f05e0e2c08a002c0450/composer.json#L23

PR #448, opened 2025-10-14
Fix `MappingDriverChain` namespace matching

HypeMC:
When I tried using multiple entity managers, I encountered an issue where the manager registry returned the wrong entity manager.

The root cause was how the `MappingDriverChain` checks whether an entity belongs to a given namespace.

In my setup, I have two namespaces: `App\Entity` and `App\EntityUser`. Since `App\EntityUser\SomeEntity` starts with `App\Entity`, it was incorrectly matched, causing the wrong entity manager to be selected.

### DIFF
--- a/tests/Persistence/Mapping/DriverChainTest.php
+++ b/tests/Persistence/Mapping/DriverChainTest.php
@@ -15,7 +15,11 @@ use stdClass;
 
 class DriverChainTest extends DoctrineTestCase
 {
-    public function testDelegateToMatchingNamespaceDriver(): void
+    /**
+     * @testWith ["Doctrine\\Tests\\Models\\Company"]
+     *           ["Doctrine\\Tests\\Persistence\\Map"]
+     */
+    public function testDelegateToMatchingNamespaceDriver(string $namespace): void
     {
         $className     = DriverChainEntity::class;
         $classMetadata = $this->createMock(ClassMetadata::class);
@@ -37,7 +41,7 @@ class DriverChainTest extends DoctrineTestCase
                 ->with(self::equalTo($className))
                 ->willReturn(true);
 
-        $chain->addDriver($driver1, 'Doctrine\Tests\Models\Company');
+        $chain->addDriver($driver1, $namespace);
         $chain->addDriver($driver2, 'Doctrine\Tests\Persistence\Mapping');
 
         $chain->loadMetadataForClass($className, $classMetadata);
--- a/src/Persistence/Mapping/Driver/MappingDriverChain.php
+++ b/src/Persistence/Mapping/Driver/MappingDriverChain.php
@@ -8,6 +8,7 @@ use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\MappingException;
 
 use function array_keys;
+use function rtrim;
 use function spl_object_hash;
 use function strpos;
 
@@ -73,7 +74,7 @@ class MappingDriverChain implements MappingDriver
     public function loadMetadataForClass(string $className, ClassMetadata $metadata)
     {
         foreach ($this->drivers as $namespace => $driver) {
-            if (strpos($className, $namespace) === 0) {
+            if ($this->isInNamespace($className, $namespace)) {
                 $driver->loadMetadataForClass($className, $metadata);
 
                 return;
@@ -105,7 +106,7 @@ class MappingDriverChain implements MappingDriver
             }
 
             foreach ($driverClasses[$oid] as $className) {
-                if (strpos($className, $namespace) !== 0) {
+                if (! $this->isInNamespace($className, $namespace)) {
                     continue;
                 }
 
@@ -128,7 +129,7 @@ class MappingDriverChain implements MappingDriver
     public function isTransient(string $className)
     {
         foreach ($this->drivers as $namespace => $driver) {
-            if (strpos($className, $namespace) === 0) {
+            if ($this->isInNamespace($className, $namespace)) {
                 return $driver->isTransient($className);
             }
         }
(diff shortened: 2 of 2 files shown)


tests/Persistence/Mapping/DriverChainTest.php
```diff
@@ -37,8 +41,8 @@ public function testDelegateToMatchingNamespaceDriver(): void
                 ->with(self::equalTo($className))
                 ->willReturn(true);
 
-        $chain->addDriver($driver1, 'Doctrine\Tests\Models\Company');
-        $chain->addDriver($driver2, 'Doctrine\Tests\Persistence\Mapping');
+        $chain->addDriver($driver1, $namespace1);
+        $chain->addDriver($driver2, $namespace2);
```
stof: Why making `$namespace2` configurable, while `$className` is not so there is not much choice for values for `$namespace2` that would not break the test ?

HypeMC: Good point, thanks. Fixed.

GromNaN (approved):

stof (approved):

greg0ire:
Thanks @HypeMC !

DarkLuk42:
**The config parameter is named "prefix", so it should be handled as a prefix, not as namespace.**
I explicitly used the prefix as a prefix and now it does not work anymore.

```yaml
entity_managers:
  em1:
    mappings:
      App\Entity:
        is_bundle: false
        dir: '%kernel.project_dir%/src/Entity'
        prefix: 'App\Entity'
```

This config, should load 'App\EntityUser...', why don't you use 'App\Entity\\' as a prefix, shouldn't it work this way?

Please consider undoing this change.

GromNaN:
`App\EntityUser` is not in the configured directory (if you follow PSR-1 or 4), so it's not loaded.

Add a 2nd mapping:
```yaml
entity_managers:
  em1:
    mappings:
      App\Entity:
        is_bundle: false
        dir: '%kernel.project_dir%/src/Entity'
        prefix: 'App\Entity'
      App\EntityUser:
        is_bundle: false
        dir: '%kernel.project_dir%/src/EntityUser'
        prefix: 'App\EntityUser'
```

DarkLuk42:
My Problem is actually a other, i have following Setup:
```yaml
entity_managers:
  em1:
    mappings:
      App\Entity:
        is_bundle: false
        dir: '%kernel.project_dir%/src/Entity'
        prefix: 'App\Entity\User'
  em2:
    mappings:
      App\Entity:
        is_bundle: false
        dir: '%kernel.project_dir%/src/Entity'
        prefix: 'App\Entity\Tenant'
 ```
 If the "prefix" option is handled as namespace, why is it not called that way?
 These are both in the same directory, and in the 3.4.1 it worked, now it is broken.

**Comment by GromNaN** [comment]
I think you're right, it's a regression. We should revert this PR. 
@HypeMC how do you configure the target directories?

**Comment by HypeMC** [comment]
@GromNaN Exactly like your example:

```yaml
entity_managers:
  em1:
    mappings:
      App\Entity:
        is_bundle: false
        dir: '%kernel.project_dir%/src/Entity'
        prefix: 'App\Entity'
  em2:
    mappings:
      App\EntityUser:
        is_bundle: false
        dir: '%kernel.project_dir%/src/EntityUser'
        prefix: 'App\EntityUser'
```

I'm a little confused about @DarkLuk42's example though, shouldn't it be:

```yaml
entity_managers:
  em1:
    mappings:
      App\Entity\User:
        is_bundle: false
        dir: '%kernel.project_dir%/src/Entity/User'
        prefix: 'App\Entity\User'
  em2:
    mappings:
      App\Entity\Tenant:
        is_bundle: false
        dir: '%kernel.project_dir%/src/Entity/Tenant'
        prefix: 'App\Entity\Tenant'
```

Also, according to the [docs](https://symfony.com/doc/current/reference/configuration/doctrine.html#prefix):

> **prefix**
> A common namespace prefix that all entities of this mapping share. This prefix should never conflict with prefixes of other defined mappings; otherwise, some of your entities cannot be found by Doctrine.

DarkLuk42:
No, in my example were multiple Entities for different entity managers in the same directory.

src/Entity/User.php
src/Entity/NotAUser.php

Before it was possible with prefix to select the User entity with prefix: 'App\Entity\U' or prefix: 'App\Entity\User'.
After this merge request it is no longer possible.
I already refactored my code and now have a separate namespaces for entities with different entity managers, so it again works with 3.4.2.
I don't need a revert/fix, but i think the old behavior to treat the prefix as prefix is the correct way, and may others have the same issue as me.

@HypeMC can't you just append a '\\' at the end of your prefix and it should work without this fix?
```yaml
entity_managers:
  em1:
    mappings:
      App\Entity:
        is_bundle: false
        dir: '%kernel.project_dir%/src/Entity'
        prefix: 'App\Entity\'
  em2:
    mappings:
      App\EntityUser:
        is_bundle: false
        dir: '%kernel.project_dir%/src/EntityUser'
        prefix: 'App\EntityUser'
```

HypeMC:
> No, in my example were multiple Entities for different entity managers in the same directory.
>
> src/Entity/User.php
> src/Entity/NotAUser.php

@DarkLuk42 Oh, thanks, I understand now. 

> can't you just append a '\' at the end of your prefix and it should work without this fix?

Yes, that would solve the issue on my side if the fix was reverted. I'm just wondering what the expected behavior is.

@GromNaN In any case, if needed I'll open a PR which reverts my fix.

GromNaN:
> In any case, if needed I'll open a PR which reverts my fix.

Yes please. We overlooked the impact.
The use-case described by @DarkLuk42 should be covered by a test.

PR #450, opened 2025-10-15
Support closure returned by `.php` files in the `PHPDriver`

GromNaN:
The [`PHPDriver`](https://github.com/doctrine/persistence/blob/eca73be1a33892e09baa6ec3d7c4fcbfd4cb5dcb/src/Persistence/Mapping/Driver/PHPDriver.php#L36-L42) use an undocumented `$metadata` variable when loading the files.
Example: https://github.com/doctrine/orm/blob/3.5.x/tests/Tests/ORM/Mapping/php/Doctrine.Tests.Models.Upsertable.Updatable.php

The idea is to accept a closure with a typed parameter, and prevent access to the loader with `$this`.

This is inspired by the way [loaders work in Symfony](https://github.com/symfony/symfony/blob/1b94288404e5555190f3c995324cfe3db6f3651d/src/Symfony/Component/Routing/Loader/PhpFileLoader.php#L46-L48).

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -16,6 +16,31 @@ The interface `Doctrine\Persistence\Mapping\ClassMetadata` has two new methods:
 
 Not implementing these methods is deprecated. They will be required in 5.0.
 
+## Deprecated modifying `$metadata` in PHP mapping files
+
+Relying on the `$metadata` variable directly in PHP mapping files is deprecated.
+Instead, wrap the code in a closure that is returned by the configuration file.
+
+Before:
+
+```php
+<?php // mappings/App.Entity.User.php
+
+$metadata->name = \App\Entity\User::class;
+```
+
+After:
+
+```php
+<?php // mappings/App.Entity.User.php
+
+use Doctrine\Persistence\Mapping\ClassMetadata;
+
+return function (ClassMetadata $metadata): void {
+    $metadata->name = \App\Entity\User::class;
+};
+```
+
 # Upgrade to 4.0
 
 ## BC Break: Removed `StaticReflectionService`
--- a/docs/en/reference/index.rst
+++ b/docs/en/reference/index.rst
@@ -258,10 +258,15 @@ mapping metadata.
 .. code-block:: php
 
     use App\Model\User;
+    use Doctrine\Persistence\Mapping\ClassMetadata;
 
-    $metadata->name = User::class;
+    return function (ClassMetadata $metadata): void {
+        $metadata->name = User::class;
 
-    // ...
+        // ...
+
+    };
+
 
 StaticPHPDriver
 --------------
@@ -283,6 +288,8 @@ Your class in ``App\Model\User`` would look like the following.
 .. code-block:: php
 
     namespace App\Model;
+
+    use Doctrine\Persistence\Mapping\ClassMetadata;
 
     final class User
     {
--- a/src/Persistence/Mapping/Driver/PHPDriver.php
+++ b/src/Persistence/Mapping/Driver/PHPDriver.php
@@ -4,7 +4,11 @@ declare(strict_types=1);
 
 namespace Doctrine\Persistence\Mapping\Driver;
 
+use Closure;
+use CompileError;
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\Persistence\Mapping\ClassMetadata;
+use Error;
 
 /**
  * The PHPDriver includes php files which just populate ClassMetadataInfo
@@ -35,6 +39,33 @@ class PHPDriver extends FileDriver
      */
     protected function loadMappingFile(string $file): array
     {
+        try {
+            $callback = Closure::bind(static function (string $file): mixed {
+                $metadata = null;
+
+                return include $file;
+            }, null, null)($file);
+        } catch (CompileError $e) {
+            throw $e;
+        } catch (Error) {
+            // Calling any method on $metadata=null will raise an Error
+            // Falling back to legacy behavior of expecting $metadata to be populated
+            $callback = null;
+        }
+
+        if ($callback instanceof Closure) {
+            $callback($this->metadata);
+
+            return [$this->metadata->getName() => $this->metadata];
+        }
+
+        Deprecation::trigger(
+            'doctrine/persistence',
+            'https://github.com/doctrine/persistence/pull/450',
+            'Not returning a Closure from a PHP mapping file is deprecated',
+        );
+
+        unset($callback);
         $metadata = $this->metadata;
         include $file;
 
--- a/tests/Persistence/Mapping/_files/Doctrine.Tests.Persistence.Mapping.PHPTestEntityClosure.php
+++ b/tests/Persistence/Mapping/_files/Doctrine.Tests.Persistence.Mapping.PHPTestEntityClosure.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+use Doctrine\Persistence\Mapping\ClassMetadata;
+
+return static function (ClassMetadata $metadata): void {
+    $metadata->getFieldNames();
+};
--- a/tests/Persistence/Mapping/_files/Doctrine.Tests.Persistence.Mapping.PHPTestIncorrectUseStatic.php
+++ b/tests/Persistence/Mapping/_files/Doctrine.Tests.Persistence.Mapping.PHPTestIncorrectUseStatic.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+use Doctrine\Persistence\Mapping\ClassMetadata;
+
+return static function (ClassMetadata $metadata): void {
+    $metadata->isIdentifier(static::class);
+};
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
     ],
     "require": {
         "php": "^8.1",
+        "doctrine/deprecations": "^1",
         "doctrine/event-manager": "^1 || ^2",
         "psr/cache": "^1.0 || ^2.0 || ^3.0"
     },
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -14,7 +14,7 @@ parameters:
         - tests/Persistence/Mapping/_files/colocated/Foo.mphp
 
     excludePaths:
-        - tests/Persistence/Mapping/_files/Doctrine.Tests.Persistence.Mapping.PHPTestEntity.php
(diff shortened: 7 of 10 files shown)


tests/Persistence/Mapping/_files/Doctrine.Tests.Persistence.Mapping.PHPTestEntityClosure.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+use Doctrine\Persistence\Mapping\ClassMetadata;
+
+return static function (ClassMetadata $metadata): void {
```
GromNaN, line 7: `static` is added by CS fix.

greg0ire:
> The [PHPDriver](https://github.com/doctrine/persistence/blob/eca73be1a33892e09baa6ec3d7c4fcbfd4cb5dcb/src/Persistence/Mapping/Driver/PHPDriver.php#L36-L42) use an undocumented $mapping variable when loading the files.
Example: https://github.com/doctrine/orm/blob/3.5.x/tests/Tests/ORM/Mapping/php/Doctrine.Tests.Models.Upsertable.Updatable.php

I don't get it; the file you mention does not contain a `$mapping` variable.

src/Persistence/Mapping/Driver/PHPDriver.php
```diff
@@ -36,7 +37,12 @@ public function loadMetadataForClass(string $className, ClassMetadata $metadata)
     protected function loadMappingFile(string $file): array
     {
         $metadata = $this->metadata;
-        include $file;
+        $result   = include $file;
+
+        if ($result instanceof Closure) {
+            $result = Closure::bind($result, null, null);
```
greg0ire: Why is this necessary?

GromNaN:
> I don't get it; the file you mention does not contain a `$mapping` variable.

Oups, I didn't write the word I was thinking of. I updated the description, it's the `$metadata` variable.

derrabus:
> I don't get it; the file you mention does not contain a `$mapping` variable.

I think he meant the `$metadata` variable.

src/Persistence/Mapping/Driver/PHPDriver.php
GromNaN: This prevent access to the `PHPDriver` instance with `$this` from the closure.

derrabus: If we `include` a PHP file from within our loader and that PHP file returns a non-static closure, that closure will be bound to the loader which means the loader and all its private variables are acessible through `$this`. By unbinding it before we execute it, we make sure that the closures can't mess with the loader.

derrabus: btw we should add a test that covers this behavior.

nicolas-grekas: this should be done around the include I think, like done here:
https://github.com/symfony/symfony/blob/1b94288404e5555190f3c995324cfe3db6f3651d/src/Symfony/Component/Routing/Loader/PhpFileLoader.php#L46-L48

tests/Persistence/Mapping/_files/Doctrine.Tests.Persistence.Mapping.PHPTestIncorrectUseStatic.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+use Doctrine\Persistence\Mapping\ClassMetadata;
+
+return static function (ClassMetadata $metadata): void {
+    // @phpstan-ignore outOfClass.static
```
derrabus: Shall we exclude the whole `_files` folder from PHPStan's analysis instead?

src/Persistence/Mapping/Driver/PHPDriver.php
GromNaN: @nicolas-grekas Indeed using the closure outside of the `include` enable usage of named closures in the included file. Although I'm not sure this is useful.

```php
return (new class () {
    private string $id = 'custom-id';

    public function __invoke(ClassMetadata $metadata): void
    {
        $metadata->isIdentifier($this->id);
    }
})(...);
```

tests/Persistence/Mapping/_files/Doctrine.Tests.Persistence.Mapping.PHPTestIncorrectUseStatic.php
GromNaN: I like this annotation that shows that the error will be reported to anyone doing this.

src/Persistence/Mapping/Driver/PHPDriver.php
```diff
@@ -35,9 +36,15 @@ public function loadMetadataForClass(string $className, ClassMetadata $metadata)
      */
     protected function loadMappingFile(string $file): array
     {
-        $metadata = $this->metadata;
```
GromNaN, line 38: To be transparent, this introduces ~2~ 1 breaking changes:
- ~the variable `$metadata` can no longer be replaced~ (not possible)
- the included field don't have access to `$this` and `self`

I would deprecate this behaviors first. And maybe deprecate not returning a closure.

derrabus: Yes to both. 💪🏻

GromNaN: The code way more verbose, but it can be greatly simplified for the next major version:
```php
protected function loadMappingFile(string $file): array
{
    $callback = Closure::bind(static function ($file): mixed {
        return include $file;
    }, null, null)($file);

    if (! $callback instanceof Closure) {
        throw new MappingException('The PHP mapping file did not return a Closure.');
    }

    $callback($this->metadata);

    return [$this->metadata->getName() => $this->metadata];
}
```

tests/Persistence/Mapping/_files/Doctrine.Tests.Persistence.Mapping.PHPTestIncorrectUseStatic.php
GromNaN: Done.

greg0ire:
This should target 4.2.x

docs/en/reference/index.rst
```diff
@@ -283,6 +288,7 @@ Your class in ``App\Model\User`` would look like the following.
 .. code-block:: php
 
     namespace App\Model;
+    use Doctrine\Persistence\Mapping\ClassMetadata;
```
greg0ire, line 292:
```suggestion

    use Doctrine\Persistence\Mapping\ClassMetadata;
```

src/Persistence/Mapping/Driver/PHPDriver.php
```diff
@@ -35,6 +39,33 @@ public function loadMetadataForClass(string $className, ClassMetadata $metadata)
      */
     protected function loadMappingFile(string $file): array
     {
+        try {
+            $callback = Closure::bind(static function ($file): mixed {
```
greg0ire:
```suggestion
            $callback = Closure::bind(static function (string $file): mixed {
```

src/Persistence/Mapping/Driver/PHPDriver.php
```diff
@@ -35,6 +39,33 @@ public function loadMetadataForClass(string $className, ClassMetadata $metadata)
...
+                $metadata = null;
+
+                return include $file;
+            }, null, null)($file);
+        } catch (CompileError $e) {
+            throw $e;
+        } catch (Error) {
+            // Accessing the undefined $metadata variable or $this causes an Error
```
greg0ire: I don't get it… when I modify this catch statement to retry the exception, here is what is printed:

```
Error: Call to a member function getFieldNames() on null
```

There is no mention of accessing a `$metadata` property on $this.

phpunit.xml.dist
```diff
@@ -5,6 +5,9 @@
          beStrictAboutOutputDuringTests="true"
          displayDetailsOnPhpunitDeprecations="true"
          displayDetailsOnTestsThatTriggerDeprecations="true"
+         displayDetailsOnTestsThatTriggerNotices="true"
+         displayDetailsOnTestsThatTriggerWarnings="true"
+         displayDetailsOnTestsThatTriggerErrors="true"
```
greg0ire: This is a welcome improvement, but if you retarget to 4.2.x, this change should still go to 4.1.x or lower.

greg0ire (commented):
Deprecations should be documented in UPGRADE.md

GromNaN:
> This should target 4.2.x

The branch 4.1 need to be upmerged into 4.2 first in order to get the PHPUnit upgrade.

greg0ire:
OK, I'm on it.

greg0ire:
Done.

phpunit.xml.dist
GromNaN: Yes. Done in https://github.com/doctrine/persistence/pull/454

src/Persistence/Mapping/Driver/PHPDriver.php
GromNaN: Exact, I added the `$metadata=null` variable after writing this comment. That's the calls to `null->method()` that results in an `Error`. Comment updated.

GromNaN:
Thanks for the review @greg0ire. The branch is rebased on 4.2 and updated.

UPGRADE.md
```diff
@@ -6,6 +6,34 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 4.2
+
+## Deprecated updating `$metadata` in PHP config files
```
greg0ire:
```suggestion
## Deprecated modifying `$metadata` in PHP mapping files
```

UPGRADE.md
```diff
@@ -6,6 +6,34 @@ awareness about deprecated code.
...
+
+Updating the `$metadata` variable directly in PHP config files is deprecated.
```
greg0ire:
```suggestion
Relying on the `$metadata` variable directly in PHP mapping files is deprecated.
```

phpunit.xml.dist
greg0ire: This should be removed from this pull request, or at least this commit, it's a concern that is orthogonal to that of the main commit.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @GromNaN !

PR #451, opened 2025-10-15
Add `getFieldValue` and `setFieldValue` to `ClassMetadata` interface

### LINKED ISSUE #426: Add `getFieldValue($object)` and `setFieldValue($object, $value)` to `ClassMetadata`
#### What

Since https://github.com/doctrine/orm/pull/11659, the way to access the value of a field has changed. This change required a patch in Symfony https://github.com/symfony/symfony/pull/59887 that then have to be patched again for ODM https://github.com/symfony/symfony/pull/60953

In order to establish a safe way to get and set the property values of an object through its class metadata, the methods `getFieldValue($object)` and `setFieldValue($object, $value)` need to be defined in the interface.

This method already exist in implementations:
- ORM: https://github.com/doctrine/orm/blob/ee919d62315680a607331d1cf0852c874362b522/src/Mapping/ClassMetadata.php#L686
- ODM: https://github.com/doctrine/mongodb-odm/blob/6f4f0fbc51e918144839de0ca5377ed5dcec3b3f/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadata.php#L1913

#### Why

Accessing a field value need to be in a common interface, that all Doctrine ORM/ODM projects can implement.

#### How

Adding a method to an interface is a breaking change; in order to introduce this new method, a new interface can be introduced.

GromNaN:
Close #426

This methods are already in all implementations of `ClassMetadata`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,16 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 4.2
+
+## Add `getFieldValue` and `setFieldValue` to `ClassMetadata` implementation
+
+The interface `Doctrine\Persistence\Mapping\ClassMetadata` has two new methods:
+- `getFieldValue(object $object, string $field)`
+- `setFieldValue(object $object, string $field, mixed $value): void`
+
+Not implementing these methods is deprecated. They will be required in 5.0.
+
 # Upgrade to 4.0
 
 ## BC Break: Removed `StaticReflectionService`
--- a/src/Persistence/Mapping/ClassMetadata.php
+++ b/src/Persistence/Mapping/ClassMetadata.php
@@ -10,6 +10,8 @@ use ReflectionClass;
  * Contract for a Doctrine persistence layer ClassMetadata class to implement.
  *
  * @template-covariant T of object
+ * @method mixed getFieldValue(object $entity, string $field)
+ * @method void  setFieldValue(object $entity, string $field, mixed $value)
  */
 interface ClassMetadata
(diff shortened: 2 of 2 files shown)


greg0ire:
This should target 4.2.x

UPGRADE.md
```diff
@@ -6,6 +6,14 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 5.0
```
greg0ire: This section should be about 4.2.0

UPGRADE.md
```diff
@@ -6,6 +6,14 @@ awareness about deprecated code.
...
+
+## BC Break: Add `getFieldValue` and `setFieldValue` to `ClassMetadata` implementation
```
greg0ire: This should be about how it's deprecated not to implement the methods you added.

GromNaN: The upgrade document is made to help people update their project. I find it more explicit to say what they have to do.

greg0ire (approved):

UPGRADE.md
greg0ire: Yes but technically, they don't _have to_ do it yet, and I have sometimes seen prophecies in minor branches become inaccurate when the change was effectively carried out on the major branch. I'll concede that in this case, it's unlikely, but let's be consistent with how we usually do this.

SenseException (approved):

greg0ire:
Thanks @GromNaN

VincentLanglet:
I think https://github.com/doctrine/persistence/issues/426 need to be closed manually @greg0ire

PR #454, opened 2025-10-16
Show more details about errors in PHPUnit

GromNaN:
Extracted from https://github.com/doctrine/persistence/pull/450#discussion_r2437331638

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -4,7 +4,12 @@
          colors="true"
          beStrictAboutOutputDuringTests="true"
          displayDetailsOnPhpunitDeprecations="true"
-         displayDetailsOnTestsThatTriggerDeprecations="true"
+         failOnDeprecation="true"
+         failOnNotice="true"
+         failOnWarning="true"
+         failOnIncomplete="true"
+         failOnRisky="true"
+         failOnEmptyTestSuite="true"
 >
     <testsuites>
         <testsuite name="Doctrine Persistence Test Suite">
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -4,6 +4,7 @@
          colors="true"
          beStrictAboutOutputDuringTests="true"
          displayDetailsOnPhpunitDeprecations="true"
+         displayDetailsOnTestsThatTriggerDeprecations="true"
          failOnDeprecation="true"
          failOnNotice="true"
          failOnWarning="true"
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -12,4 +12,4 @@ on:
 jobs:
   coding-standards:
     name: "Coding Standards"
-    uses: "doctrine/.github/.github/workflows/coding-standards.yml@12.0.0"
+    uses: "doctrine/.github/.github/workflows/coding-standards.yml@12.1.0"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -12,8 +12,9 @@ on:
 jobs:
   phpunit:
     name: "PHPUnit"
-    uses: "doctrine/.github/.github/workflows/continuous-integration.yml@12.0.0"
+    uses: "doctrine/.github/.github/workflows/continuous-integration.yml@12.1.0"
     with:
       php-versions: '["8.1", "8.2", "8.3", "8.4", "8.5"]'
+      phpunit-options-lowest: "--do-not-fail-on-deprecation"
     secrets:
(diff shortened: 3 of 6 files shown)


phpunit.xml.dist
```diff
@@ -5,6 +5,9 @@
          beStrictAboutOutputDuringTests="true"
          displayDetailsOnPhpunitDeprecations="true"
          displayDetailsOnTestsThatTriggerDeprecations="true"
+         displayDetailsOnTestsThatTriggerNotices="true"
+         displayDetailsOnTestsThatTriggerWarnings="true"
+         displayDetailsOnTestsThatTriggerErrors="true"
```
greg0ire: Should we maybe use the `failOn` counterparts here?

phpunit.xml.dist
```diff
@@ -5,6 +5,11 @@
...
+         failOnNotice="true"
+         failOnWarning="true"
```
greg0ire, line 10: I would expect `failOnNotice` to imply `displayDetailsOnTestsThatTriggerNotices`. Can you check if that is the case?

GromNaN: You are right. With `failOnWarning`, the details are displayed.

phpunit.xml.dist
```diff
@@ -4,7 +4,12 @@
          colors="true"
          beStrictAboutOutputDuringTests="true"
          displayDetailsOnPhpunitDeprecations="true"
-         displayDetailsOnTestsThatTriggerDeprecations="true"
+         displayDetailsOnTestsThatTriggerErrors="true"
```
greg0ire: Regarding `displayDetailsOnTestsThatTriggerErrors`: what's the difference between having it and the default behavior?

greg0ire: @GromNaN I tried `throw \Error` inside a test, and couldn't find a difference with or without this setting, so let's remove it?

GromNaN: The different is with `trigger_error('Hello World', E_USER_ERROR)`.

```diff
  PHPUnit 12.4.1 by Sebastian Bergmann and contributors.
  
  Runtime:       PHP 8.4.8
  Configuration: /Users/jerome/Develop/doctrine-persistence/phpunit.xml.dist

  E                                                                   1 / 1 (100%)

  Time: 00:00.006, Memory: 16.00 MB

  There was 1 error:
  
  1) ErrorTest::testError
  E_USER_ERROR was triggered
  tests/ErrorTest.php:9

+ --
+ 
+ 1 test triggered 1 error:
+ 
+ 1) tests/ErrorTest.php:9
+ Hello World
+
  ERRORS!
  Tests: 1, Assertions: 0, Errors: 2, Deprecations: 1.
```

greg0ire: Ah nice! I guess it could be useful to have the "Hello World" part.

greg0ire (dismissed):

SenseException (dismissed):

phpunit.xml.dist
```diff
@@ -4,7 +4,12 @@
          colors="true"
          beStrictAboutOutputDuringTests="true"
          displayDetailsOnPhpunitDeprecations="true"
-         displayDetailsOnTestsThatTriggerDeprecations="true"
```
xabbuh, line 7: There is now no entry for deprecations anymore. Should `failOnDeprecation` be added as a replacement?

phpunit.xml.dist
GromNaN: This was my test:
```php
<?php

use PHPUnit\Framework\TestCase;

class ErrorTest extends TestCase
{
    public function testError(): void
    {
        trigger_error('Hello World', E_USER_ERROR);
    }
}
```

GromNaN:
We get deprecation with PHP 8.1, without message
```
PHPUnit 10.5.58 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.33 with PCOV 1.0.12
Configuration: /home/runner/work/persistence/persistence/phpunit.xml.dist

...................................................DS..........  63 / 137 ( 45%)
............................................................... 126 / 137 ( 91%)
...........                                                     137 / 137 (100%)

Time: 00:00.258, Memory: 22.00 MB

OK, but there were issues!
Tests: 137, Assertions: 249, Deprecations: 10, Skipped: 1.

Generating code coverage report in Clover XML format ... done [00:00.020]
```

xabbuh:
When the job uses lowest dependencies, PHPUnit should probably be instructed to not fail when detecting deprecations using the `--do-not-fail-on-deprecation` flag.

GromNaN:
I updated the Action. PR is ready.

greg0ire (approved):
👌

PR #458, opened 2025-10-29
Make `AbstractManagerRegistry::$proxyInterfaceName` nullable

GromNaN:
When native lazy objects are used, there is no proxy class wrapping the entity/document class.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -41,6 +41,11 @@ return function (ClassMetadata $metadata): void {
 };
 ```
 
+## Do not pass any proxy interface to `AbstractManagerRegistry` when using native proxies
+
+With PHP 8.4 native lazy objects, you don't need to pass any proxy interface to
+`AbstractManagerRegistry`. The class of the lazy objects is the class being mapped.
+
 # Upgrade to 4.0
 
 ## BC Break: Removed `StaticReflectionService`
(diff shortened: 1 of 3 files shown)


src/Persistence/AbstractManagerRegistry.php
```diff
@@ -47,7 +47,7 @@ public function __construct(
         array $managers,
         string $defaultConnection,
         string $defaultManager,
-        string $proxyInterfaceName
+        ?string $proxyInterfaceName
```
stof: the phpdoc type should be updated as well.

And the parameter should probably default to `null` to allow removing it in the next major version of `doctrine/persistence` in a future where proxies are gone entirely.

greg0ire (dismissed):

SenseException (dismissed):

greg0ire:
@GromNaN why are you targeting 3.4.x ? Is this a bugfix?

GromNaN:
It's not absolutely a bug since you can pass the `Doctrine\Persistence\Proxy` interface name instead even if native lazy objects don't implement it. But the name of an existing interface is required.

greg0ire:
That's not a bug indeed, that's a design issue. I think you should target 4.2.x

greg0ire:
Or rather, just close in favor of https://github.com/doctrine/persistence/pull/460

GromNaN:
I rebased this PR on 4.2, since #460 is going to be for 5.x.

Skipping the `ReflectionClass` is a small performance optimization.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @GromNaN !

PR #460, opened 2025-10-30
Deprecate `Proxy` interface and related

GromNaN:
Native lazy objects don't need proxy classes. By deprecating interfaces and methods related to proxies, we can spot them more easily in integration and ensure we don't use them when native lazy objects are enabled.

### DIFF
--- a/src/AbstractManagerRegistry.php
+++ b/src/AbstractManagerRegistry.php
@@ -18,7 +18,7 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
     /**
      * @param array<string, string> $connections
      * @param array<string, string> $managers
-     * @phpstan-param class-string $proxyInterfaceName
+     * @phpstan-param class-string|null $proxyInterfaceName
      */
     public function __construct(
         private readonly string $name,
@@ -26,7 +26,7 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
         private array $managers,
         private readonly string $defaultConnection,
         private readonly string $defaultManager,
-        private readonly string $proxyInterfaceName,
+        private readonly string|null $proxyInterfaceName = null,
     ) {
     }
 
@@ -132,7 +132,7 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
             return null;
         }
 
-        if ($proxyClass->implementsInterface($this->proxyInterfaceName)) {
+        if ($this->proxyInterfaceName !== null && $proxyClass->implementsInterface($this->proxyInterfaceName)) {
             $parentClass = $proxyClass->getParentClass();
 
             if ($parentClass === false) {
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,13 @@ awareness about deprecated code.
 
 # Upgrade to 5.0
 
+## Deprecate `Proxy` and `ProxyClassNameResolver` interfaces
+
+Use native lazy objects instead of proxy classes for lazy objects.
+Don't use the `Doctrine\Persistence\Proxy` interface to identify lazy objects.
+Remove implementations and usages of the `ProxyClassNameResolver`. Native lazy
+objects have the same class.
+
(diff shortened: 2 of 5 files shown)


src/AbstractManagerRegistry.php
```diff
@@ -18,15 +18,15 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
...
+     * @phpstan-param class-string|null $proxyInterfaceName
      */
     public function __construct(
         private readonly string $name,
         private array $connections,
         private array $managers,
         private readonly string $defaultConnection,
         private readonly string $defaultManager,
-        private readonly string $proxyInterfaceName,
+        private readonly string|null $proxyInterfaceName = null,
```
GromNaN, line 29: - Also in 3.4: https://github.com/doctrine/persistence/pull/458

src/Persistence/Mapping/AbstractClassMetadataFactory.php
```diff
@@ -232,8 +233,6 @@ public function hasMetadataFor(string $className): bool
     /**
      * Sets the metadata descriptor for a specific class.
      *
-     * NOTE: This is only useful in very special cases, like when generating proxy classes.
-     *
      * @phpstan-param class-string $className
      * @phpstan-param CMTemplate $class
      */
```
GromNaN, line 239: This method is never used; last usage removed by
- https://github.com/doctrine/orm/pull/12174

derrabus (dismissed):

stof:
Deprecating those seems too early when the package (and the object manager packages implementing the interface) still have actively maintained versions that support PHP 8.3 and older (and so cannot use native lazy objects unconditionally)

stof:
For reference, the ORM triggers its own deprecation warning about using proxies instead of native lazy objects only when running on PHP 8.4+

GromNaN:
> Deprecating those seems too early when the package (and the object manager packages implementing the interface) still have actively maintained versions that support PHP 8.3 and older (and so cannot use native lazy objects unconditionally)

That's what I think too. Deprecations are not triggered directly, but the autoloader will do it. It's mainly to identify classes and interfaces that should no longer be used when native lazy is enabled.

greg0ire:
I think it might make sense to say that 4.2.x will be the last 4.x, and to bump to 8.4 so that it's fully OK to have these deprecations. We may even release 5.0.0 before v4 of the ORM, and have ORM v3 be compatible with persistence v5 (and it would crash if you would attempt to use the legacy proxies with persistence v5). Then, on orm v4, we drop compatibility with persistence v4. The upgrade path for people would be:

1. upgrade to ORM 3.4
2. upgrade to PHP 8.4
1. enable lazy objects
3. upgrade to persistence v5
4. upgrade to ORM v4

stof:
> I think it might make sense to say that 4.2.x will be the last 4.x, and to bump to 8.4 so that it's fully OK to have these deprecations.

this would imply making 4.1.x a maintained version that could receive patches if needed, unless ORM and ODM both drop support for PHP 8.3 first (so that users of a maintained ORM version on PHP 8.3 are not forced to install an unmaintained persistence version.

greg0ire:
Ah yes that's not too great. Maybe this needs to be postponed to 5.1.x then

greg0ire (approved):
I wrote 5.1 but it does not exist yet, and I think starting a new major with deprecations might be OK in this case.

GromNaN:
> I wrote 5.1 but it does not exist yet, and I think starting a new major with deprecations might be OK in this case.

I also vote for 5.0.0, otherwise we will just release 5.1.0 right after 5.0.0 and the result will be exactly the same.

SenseException (approved):

greg0ire:
@GromNaN the difference if you do that is that's it's easier to check that all deprecations that are supposed to be removed are removed before releasing 5. But no big deal anyway

greg0ire (approved):

greg0ire:
Thanks @GromNaN !

PR #473, opened 2025-11-02
Require PHP 8.4

greg0ire:
This is consistent with other Doctrine projects. Note that I had to add a "covariant" keyword on the template definition of `getReflectionClass()` return type. That's because since PHP 8.4, `ReflectionClass` is considered invariant because of lazy objects. Lazy objects introduce new methods to `ReflectionClass` where the template appears in parameter position, for instance that is the case for `isUninitializedLazyObject()`. I do not understand why this wasn't an issue with
`ReflectionClass::isInstance()`, which was introduced way earlier than that though.

The solution I picked is to use "call-site variance", effectively saying that this is OK in this instance (but it might not be, I'm not sure).

The alternative to adding this keyword would be to drop generic information entirely.

See https://github.com/phpstan/phpstan/issues/12459 for more details.

### DIFF
--- a/src/Proxy.php
+++ b/src/Proxy.php
@@ -14,12 +14,12 @@ interface Proxy
     /**
      * Marker for Proxy class names.
      */
-    public const MARKER = '__CG__';
+    public const string MARKER = '__CG__';
 
     /**
      * Length of the proxy marker.
      */
-    public const MARKER_LENGTH = 6;
+    public const int MARKER_LENGTH = 6;
 
     /**
      * Initializes this proxy if its not yet initialized.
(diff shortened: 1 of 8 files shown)


src/Proxy.php
```diff
@@ -14,12 +14,12 @@ interface Proxy
     /**
      * Marker for Proxy class names.
      */
-    public const MARKER = '__CG__';
+    public const string MARKER = '__CG__';
```
GromNaN, line 17: This interface will be deprecated. I'm not sure it's worth having a breaking change.

greg0ire: I don't think that interface is actually extended anywhere, which makes this breaking change very mild (and also, its' really easy to address).

SenseException (approved):

PR #477, opened 2025-11-24
Test with Symfony 8

loic425:


GromNaN (approved):
Thanks. This PR should be titled "Test with Symfony 8", as there is no fix and the Symfony deprendencies are optional.

loic425:
> Thanks. This PR should be titled "Test with Symfony 8", as there is no fix and the Symfony deprendencies are optional.

Yes, I'm surprised that the CI does not contain tests with several Symfony versions.

derrabus (approved):

PR #486, opened 2025-12-06
Support ClassLocator in StaticPHPDriver

GromNaN:
The `ClassLocator` abstraction added in https://github.com/doctrine/persistence/pull/433 should be used for the `StaticPHPDriver`.

By using the `ColocatedMappingDriver` trait, we remove code duplication.

- fix `StaticPHPDriverTest::testGetAllClassNames` by only scanning `/_files/colocated/`. Otherwise it loads the `Doctrine.Tests.Persistence.Mapping.PHPTestEntityAssert.php` file that expects the variable `$metadata` to be defined. This test passes when all the test suite is run, but not individually.

- Relax PHPStan type for list of directories in `FileClassLocator::createFromDirectories()` as we don't care about the array key.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -61,6 +61,17 @@ return function (ClassMetadata $metadata): void {
 };
 ```
 
+## New methods in `StaticPHPDriver`
+
+The `StaticPHPDriver` get new method to configure the scanned directories:
+- `addExcludePaths(array $paths): void`
+- `getExcludePaths(): array`
+- `setFileExtension(string $fileExtension): void`
+- `getFileExtension(): string`
+
+Using the a `ClassLocator` implementation is recommended instead of relying
+on directory scanning.
+
 ## Do not pass any proxy interface to `AbstractManagerRegistry` when using native proxies
 
 With PHP 8.4 native lazy objects, you don't need to pass any proxy interface to
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -61,15 +61,16 @@ return function (ClassMetadata $metadata): void {
 };
 ```
 
-## New methods in `StaticPHPDriver`
+## `StaticPHPDriver` now accepts a `ClassLocator`
 
-The `StaticPHPDriver` get new method to configure the scanned directories:
-- `addExcludePaths(array $paths): void`
-- `getExcludePaths(): array`
-- `setFileExtension(string $fileExtension): void`
-- `getFileExtension(): string`
+The constructor of `StaticPHPDriver` now accepts a `ClassLocator` instance
+in addition to a path or array of paths:
 
-Using the a `ClassLocator` implementation is recommended instead of relying
+```php
+$driver = new StaticPHPDriver(new ClassNames([MyEntity::class, AnotherEntity::class]));
+```
+
+Using a `ClassLocator` implementation is recommended instead of relying
 on directory scanning.
 
 ## Do not pass any proxy interface to `AbstractManagerRegistry` when using native proxies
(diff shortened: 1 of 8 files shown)


tests/Mapping/StaticPHPDriverTest.php
```diff
@@ -15,24 +17,23 @@ public function testLoadMetadata(): void
...
 
-        $driver = new StaticPHPDriver([__DIR__]);
-        $driver->loadMetadataForClass(TestEntity::class, $metadata);
+        $driver = new StaticPHPDriver([]);
+        $driver->loadMetadataForClass(Entity::class, $metadata);
     }
 
     public function testGetAllClassNames(): void
     {
-        $driver     = new StaticPHPDriver([__DIR__]);
```
GromNaN, line 24: This directory contains `Doctrine.Tests.Persistence.Mapping.PHPTestEntityAssert.php` that cannot be included without setting the `$metadata` variable. Not supported by the `StaticPHPDriver`.

src/Mapping/Driver/FileClassLocator.php
```diff
@@ -81,9 +81,9 @@ public function getClassNames(): array
     /**
      * Creates a FileClassLocator from an array of directories.
      *
-     * @param list<string> $directories
-     * @param list<string> $excludedDirectories Directories to exclude from the search.
-     * @param string       $fileExtension       The file extension to look for (default is '.php').
+     * @param string[] $directories
+     * @param string[] $excludedDirectories Directories to exclude from the search.
```
GromNaN, line 85: We don't use the key of the array when iterating. Relaxing this type so that `array<int, string>` is accepted, which is the result of `array_unique`.

src/Mapping/Driver/StaticPHPDriver.php
```diff
@@ -5,115 +5,32 @@
...
 
     public function loadMetadataForClass(string $className, ClassMetadata $metadata): void
     {
         $className::loadMetadata($metadata);
     }
 
-    /**
-     * {@inheritDoc}
-     *
-     * @todo Same code exists in ColocatedMappingDriver, should we re-use it
```
GromNaN, line 63: This deleted code is duplicate of [`FileClassLocator`](https://github.com/doctrine/persistence/blob/4.1.x/src/Mapping/Driver/FileClassLocator.php).

src/Mapping/Driver/StaticPHPDriver.php
```diff
@@ -5,115 +5,32 @@
...
  */
 class StaticPHPDriver implements MappingDriver
 {
-    /**
-     * Paths of entity directories.
-     *
-     * @var array<int, string>
-     */
-    private array $paths = [];
+    use ColocatedMappingDriver;
```
SenseException: This introduces more public methods to the class. Maybe it needs to be documented in UPGRADE.md

SenseException (changes requested):

UPGRADE.md
```diff
@@ -41,6 +41,17 @@ return function (ClassMetadata $metadata): void {
...
 
+## New methods in `StaticPHPDriver`
+
+The `StaticPHPDriver` get new method to configure the scanned directories:
+- `addExcludePaths(array $paths): void`
+- `getExcludePaths(): array`
+- `setFileExtension(string $fileExtension): void`
+- `getFileExtension(): string`
+
+Using the a `ClassLocator` implementation is recommended instead of relying
```
SenseException:
```suggestion
Using a `ClassLocator` implementation is recommended instead of relying
```

SenseException (changes requested):

src/Mapping/Driver/StaticPHPDriver.php
GromNaN: I removed the new public methods by making them private.

SenseException (approved):

greg0ire (approved):

greg0ire:
Thanks @GromNaN !

PR #489, opened 2026-01-15
Mark classes as soft final

greg0ire:


src/Reflection/RuntimeReflectionProperty.php
```diff
@@ -15,6 +15,8 @@
  *
  * Avoids triggering lazy loading if the provided object
  * is a {@see \Doctrine\Persistence\Proxy}.
+ *
+ * @final since 4.2
```
GromNaN: PHPStan spotted this class as being extended by `TypedNoDefaultReflectionProperty`. It can't be final.

UPGRADE.md
```diff
@@ -16,6 +16,28 @@ The interface `Doctrine\Persistence\Mapping\ClassMetadata` has two new methods:
...
+- `Doctrine\Persistence\Event\PreUpdateEventArgs`
+- `Doctrine\Persistence\Mapping\Driver\DefaultFileLocator`
+- `Doctrine\Persistence\Mapping\Driver\MappingDriverChain`
+- `Doctrine\Persistence\Mapping\Driver\PHPDriver`
+- `Doctrine\Persistence\Mapping\Driver\StaticPHPDriver`
+- `Doctrine\Persistence\Mapping\Driver\SymfonyFileLocator`
+- `Doctrine\Persistence\Mapping\MappingException`
+- `Doctrine\Persistence\Mapping\RuntimeReflectionService`
+- `Doctrine\Persistence\Reflection\EnumReflectionProperty`
+- `Doctrine\Persistence\Reflection\RuntimeReflectionProperty`
```
GromNaN:
```suggestion
```

GromNaN (dismissed):

src/Event/OnClearEventArgs.php
```diff
@@ -11,6 +11,7 @@
  * Provides event arguments for the onClear event.
  *
  * @template-covariant TObjectManager of ObjectManager
+ * @final since 4.2
```
stof: All those EventArgs classes are extended in the ORM and the ODM. This must be solved before we mark them as `@final` here (or we should not make them soft final)

src/Mapping/MappingException.php
```diff
@@ -11,6 +11,8 @@
 
 /**
  * A MappingException indicates that something is wrong with the mapping setup.
+ *
+ * @final since 4.2
```
stof: should this be final ? Couldn't the ORM or the ODM have valid use cases for extending it for other kind of mapping exceptions that are specific to that project ? I'm not sure whether they do it today, or if they instantiate MappingException directly or if they throw a separate type of exception for those cases.

src/Reflection/RuntimeReflectionProperty.php
stof: Wea solution could be to use the `@phpstan-sealed` or `@psalm-inheritors` property to restrict the supported child classes

As of today, phpstan is able to read the psalm annotation but [Psalm does not support reading the phpstan annotation yet](https://github.com/vimeo/psalm/issues/11510) so using the psalm annotation is needed for interoperable support for downstream projects (or contributing to Psalm to implement that feature request). Note also that phpstan currently has an [issue when using the annotation on a concrete class](https://github.com/phpstan/phpstan/issues/13734) (which is the case here). I have an open PR fixing this but it hasn't been reviewed yet. Edit: my PR is now merged and should be in the next 2.1.x release.

src/Event/OnClearEventArgs.php
greg0ire: Oh right, I grepped for `extends OnClear`, but the class was aliased. I will revert the change on these classes.

src/Mapping/MappingException.php
greg0ire: I will revert this as well, I failed to find a current extension for the same reason (aliasing)

src/Reflection/RuntimeReflectionProperty.php
```diff
@@ -15,6 +15,8 @@
  *
  * Avoids triggering lazy loading if the provided object
  * is a {@see \Doctrine\Persistence\Proxy}.
+ *
+ * @phpstan-sealed TypedNoDefaultReflectionProperty
```
greg0ire, line 19: @stof PHPStan is currently pinned to 2.1.30, however I did not encounter the issue you've mentioned. And the annotation works, I've confirmed that by using `EnumReflectionProperty`:

```
 ------ ---------------------------------------------------------------------------------------------------------
  Line   src/Reflection/TypedNoDefaultReflectionProperty.php
 ------ ---------------------------------------------------------------------------------------------------------
  16     Type Doctrine\Persistence\Reflection\TypedNoDefaultReflectionProperty is not allowed to be a subtype of
         Doctrine\Persistence\Reflection\RuntimeReflectionProperty.
         🪪  class.disallowedSubtype
 ------ ---------------------------------------------------------------------------------------------------------
```

stof: the issue I've mentioned is not related to the case of creating child classes. It happens when refining types (which might not affect this package depending on how it consumes that class)

stof (approved):

SenseException (approved):

PR #492, opened 2026-01-21
Modernize the codebase

greg0ire:


src/AbstractManagerRegistry.php
```diff
@@ -22,8 +22,8 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
      */
     public function __construct(
         private readonly string $name,
-        private array $connections,
-        private array $managers,
+        private readonly array $connections,
+        private readonly array $managers,
```
greg0ire, line 27: From my first commit message:

> It is no longer clear to me why I did not make the
AbstractManagerRegistry::$connections and
AbstractManagerRegistry::$managers readonly already in
https://github.com/greg0ire/persistence/commit/59029afa2df7cf689e21b78183e4903b53e170b8.

If you can imagine a reason why I didn't, let me know. If not, then I guess the best way to find out will be to just do it.

src/Mapping/Driver/DefaultFileLocator.php
GromNaN: The `@final` annotation is duplicating the hard `final`.

src/AbstractManagerRegistry.php
```diff
@@ -165,6 +174,7 @@ public function getManagerNames(): array
     /**
      * {@inheritDoc}
```
stof, line 175: should be removed. The inline phpdoc tag is only about inheriting the long description (which is only useful when wanting to include it inside the long description define of the child, thing about it as the `parent::` call for the phpdoc long description).
The current usage is likely a confusion with the block-level tag `@inheritDoc` that is meant to say "this method is overriding the parent one which already defines everything so don't complain about missing phpdoc". And the `#[Override]` already provides that intent.

greg0ire: @stof should we wait on https://github.com/slevomat/coding-standard/issues/1824 before attempting to address this?

src/Mapping/Driver/DefaultFileLocator.php
greg0ire: Good catch, fixed!

src/AbstractManagerRegistry.php
GromNaN: Thanks for the PR https://github.com/slevomat/coding-standard/pull/1828 @greg0ire. We can remove the phpdoc in all the doctrine projects in an other PR.

GromNaN (approved):

PR #493, opened 2026-01-24
Use Rector on tests

greg0ire:
This is the configuration I used:

```php
<?php

declare(strict_types=1);

use Rector\Config\RectorConfig;

return RectorConfig::configure()
	->withPaths([
	__DIR__ . '/tests',
	])
	->withPhpSets();
```


greg0ire:
I plan to follow this up with similar MRs on the minor and major branches.

SenseException (approved):

PR #498, opened 2026-02-04
findAll and findBy should return list

stchr:
https://github.com/doctrine/orm/blob/3.6.x/src/EntityRepository.php already has this narrowed return type

### DIFF
--- a/src/ObjectRepository.php
+++ b/src/ObjectRepository.php
@@ -27,7 +27,7 @@ interface ObjectRepository
      * Finds all objects in the repository.
      *
      * @return array<int, object> The objects.
-     * @phpstan-return <T>
+     * @phpstan-return list<T>
      */
     public function findAll(): array;
 
--- a/src/ObjectRepository.php
+++ b/src/ObjectRepository.php
@@ -27,7 +27,7 @@ interface ObjectRepository
      * Finds all objects in the repository.
      *
      * @return array<int, object> The objects.
-     * @phpstan-return T[]
+     * @phpstan-return <T>
      */
     public function findAll(): array;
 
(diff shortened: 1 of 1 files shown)


greg0ire:
This kind of change should target 4.2.x

GromNaN:
> This kind of change should target 4.2.x

I agree.

This is an additional constraint added to this methods. Do all implementations in Doctrine projects return a list?

The `list` type must be also in [Doctrine ODM `EntityPersister::loadAll()`](https://github.com/doctrine/orm/blob/4262eb495b4d2a53b45de1ac58881e0091f2970f/src/Persisters/Entity/EntityPersister.php#L229-L231C21) that is called by [`EntityRepository::findBy()`](https://github.com/doctrine/orm/blob/4262eb495b4d2a53b45de1ac58881e0091f2970f/src/EntityRepository.php#L110)

This needs to be explored, it might be an indexed array in certain conditions?
https://github.com/doctrine/orm/blob/4262eb495b4d2a53b45de1ac58881e0091f2970f/src/Internal/Hydration/ObjectHydrator.php#L508

stof:
@GromNaN the repository methods don't allow asking for indexed results (the ObjectHydrator indeed does not always return a list, it depends on the result set mapping).

greg0ire:
Then I guess the ORM types should be fixed? If yes, targeting the patch branch of ORM would be the correct thing to do.

stof:
@greg0ire why ? The `findAll` and `findBy` methods of the repository always return a list (as they don't allow requesting indexed results). The ObjectHydrator is not documented as returning a list in the ORM.

greg0ire:
Sorry, I got confused, ~I thought you were implying the hydrator and the entity methods were connected indirectly, and did not check that.~ Or maybe they are connected, but the way one uses the other, a list is always returned?

stchr:
Thanks for all the input, highly appreciated!
To summarize if I understand correctly: everything is fine, just need to target 4.2.x branch, right?

stchr:
closed in favor of #499, thanks a lot!

PR #499, opened 2026-02-05
findAll and findBy should return list

stchr:
https://github.com/doctrine/orm/blob/3.6.x/src/EntityRepository.php already has this narrowed return type

See discussion in #498


GromNaN (approved):

greg0ire:
Thanks @stchr !

PR #513, opened 2026-04-27
Convert ClassMetadata::getFieldValue/setFieldValue from @method to real interface methods

GromNaN:
Closes #451

`@param T $entity` is used on both methods even though `T` is declared covariant. This is intentional( the methods only make sense for the mapped entity type) but it violates strict generic variance rules. PHPStan warnings are suppressed with `@phpstan-ignore generics.variance`.

Open question: should `@template-covariant T` be relaxed to `@template T` (invariant) to model this more accurately? That would be a breaking change for downstream type annotations.

### DIFF
--- a/src/Mapping/ClassMetadata.php
+++ b/src/Mapping/ClassMetadata.php
@@ -4,14 +4,13 @@ declare(strict_types=1);
 
 namespace Doctrine\Persistence\Mapping;
 
+use InvalidArgumentException;
 use ReflectionClass;
 
 /**
  * Contract for a Doctrine persistence layer ClassMetadata class to implement.
  *
  * @template-covariant T of object
- * @method mixed getFieldValue(object $entity, string $field)
- * @method void  setFieldValue(object $entity, string $field, mixed $value)
  */
 interface ClassMetadata
 {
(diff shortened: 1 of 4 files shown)


src/Mapping/ClassMetadata.php
```diff
@@ -63,6 +61,24 @@ public function isCollectionValuedAssociation(string $fieldName): bool;
      */
     public function getFieldNames(): array;
 
+    /**
+     * Gets the value of the given field of the given object.
+     *
+     * @param T $entity
+     *
+     * @phpstan-ignore generics.variance (T in parameter position is intentional: the method only makes sense for the mapped entity type)
```
stof: T in parameter position requires the generic to be invariant. So we should indeed change it.

src/Mapping/ClassMetadata.php
```diff
@@ -63,6 +62,20 @@ public function isCollectionValuedAssociation(string $fieldName): bool;
      */
     public function getFieldNames(): array;
 
+    /**
+     * Gets the value of the given field of the given object.
+     *
+     * @throws InvalidArgumentException if the object is not supported, or the field is not mapped.
```
GromNaN, line 68: I also added `@throws InvalidArgumentException` to document the contract when `$object` is not of the expected type or `$field` is not mapped. Currently the ORM and ODM implementations don't throw an exception, but that would be an improvement when upgrading to persistence v5

src/Mapping/ClassMetadata.php
GromNaN: The parameter `$object` is typed as `object` rather than `T`. The interface declares `@template-covariant T of object`, meaning `ClassMetadata<Cat>` is a subtype of `ClassMetadata<Animal>`. Covariance requires T to appear only in return (output) positions. Using `@param T $object` would place T in a parameter position, violating that constraint.

I can't change to `@template T`, it's creating more issues with the child classes.

GromNaN (commented):
Now I understand how covariance works for this class.

greg0ire (approved):

PR #514, opened 2026-06-12
fix: Add parent constructor call in EnumReflectionProperty

jawira:
Hello, this PR is a backport of #422 to the 3.4.x branch.  
The original fix was merged into 4.0.x, this PR applies the same solution to the 3.4.x branch.

The issue this PR solves is explained in: https://github.com/doctrine/persistence/issues/423

### DIFF
--- a/src/Reflection/EnumReflectionProperty.php
+++ b/src/Reflection/EnumReflectionProperty.php
@@ -5,9 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Persistence\Reflection;
 
 use BackedEnum;
-use ReflectionClass;
 use ReflectionProperty;
-use ReflectionType;
 use ReturnTypeWillChange;
 
 use function array_map;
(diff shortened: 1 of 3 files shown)


src/Reflection/EnumReflectionProperty.php
```diff
@@ -71,7 +44,7 @@ public function getValue($object = null)
             return null;
         }
 
-        $enum = $this->originalReflectionProperty->getValue($object);
+        $enum = parent::getValue($object);
```
greg0ire: This does diff not appear in #422

greg0ire (commented):
Why did you not also backport the new tests?

src/Reflection/EnumReflectionProperty.php
jawira: You are right. While working in this PR, I realised the "`originalReflectionProperty`" property isn't actually needed, since a `parent` call can be used instead.  
My idea was to make this change in the "`3.4.x`" branch and then update the "`4.2.x`" branch after this PR was merged.

Would you prefer that I keep the "`originalReflectionProperty`" calls to remain consistent with the "`4.2.x`" branch, or should I proceed with the cleanup in this PR?  
Please let me know WDYT

jawira:
> Why did you not also backport the new tests?

It was an oversight, I have just commit the tests.

src/Reflection/EnumReflectionProperty.php
greg0ire: > My idea was to make this change in the "3.4.x" branch and then update the "4.2.x" branch after this PR was merged.

It's not a bugfix though, right? If not, it shouldn't go to either branch, but it should rather go to 4.3.x.

jawira: Hello, so I have reverted the two occurences of "parent::" as requested.  
It should be good now :)

greg0ire (approved):

greg0ire:
Thanks @jawira !

src/Reflection/EnumReflectionProperty.php
```diff
@@ -28,33 +26,12 @@ class EnumReflectionProperty extends ReflectionProperty
     /** @param class-string<BackedEnum> $enumType */
     public function __construct(ReflectionProperty $originalReflectionProperty, string $enumType)
     {
+        parent::__construct($originalReflectionProperty->class, $originalReflectionProperty->name);
+
         $this->originalReflectionProperty = $originalReflectionProperty;
         $this->enumType                   = $enumType;
```
greg0ire, line 32: @jawira shouldn't these 2 lines have vanished? They are not present on 4.2.x

jawira: @greg0ire Those two lines are present in the `4.2.x` branch, but they are implemented using _constructor property promotion_.

image

I can't use the same syntax in the `3.4.x` branch because _constructor property promotion_ was introduced in `PHP 8.0`, while the minimum supported PHP version for `3.4.x` is `PHP 7.2`. The syntax is different, but the semantics are the same.

greg0ire: Ah right! Didn't spot that.

PR #516, opened 2026-06-17
Refactor EnumReflectionProperty to remove $originalReflectionProperty

jawira:
Hello, in this PR I propose to remove `$originalReflectionProperty` property from `EnumReflectionProperty` class. This is a refactoring, so no new functionality has been removed or added.

The reasons are:

* This will eliminate redundant state. `EnumReflectionProperty` extends `ReflectionProperty` and initializes the parent with the same information. Keeping a second reference means the object stores the same conceptual state twice.
* Removing `$originalReflectionProperty` makes it clear that `EnumReflectionProperty` itself is the authoritative reflection object. Developers no longer need to wonder whether a method should be called on `$this`, `$this->originalReflectionProperty`, or whether the two can diverge.

### DIFF
--- a/src/Reflection/EnumReflectionProperty.php
+++ b/src/Reflection/EnumReflectionProperty.php
@@ -19,7 +19,7 @@ use function reset;
 class EnumReflectionProperty extends ReflectionProperty
 {
     /** @param class-string<BackedEnum> $enumType */
-    public function __construct(private readonly ReflectionProperty $originalReflectionProperty, private readonly string $enumType)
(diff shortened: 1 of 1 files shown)


greg0ire (dismissed):

greg0ire:
Sorry if I was unclear, but [when I ask you to target 4.3.x](https://github.com/doctrine/persistence/pull/514#discussion_r3407857303), it means you should target 4.3.x :stuck_out_tongue:

SenseException (approved):

PR #519, opened 2026-06-18
Stop maintaining 3.x

greg0ire:
The downloads of 4.x have surpassed the downloads of 3.x.

image

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -34,9 +34,8 @@
         },
         {
             "name": "3.4",
-            "branchName": "3.4.x",
             "slug": "3.4",
-            "maintained": true
+            "maintained": false
         },
         {
             "name": "3.3",
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,6 @@ updates:
       interval: "weekly"
     labels:
       - "CI"
-    target-branch: "3.4.x"
     groups:
       doctrine:
         patterns:


GromNaN:
From a [quick search](https://github.com/search?q=org%3Adoctrine+doctrine%2Fpersistence+language%3AJSON&type=code&l=JSON), I found the following packages that are not compatible with v4 yet:
- DoctrineORMModule: [requires `^3.0.0`](https://github.com/doctrine/DoctrineORMModule/blob/77a395980838ff11001dcb272b9c9b1f97c86ce5/composer.json#L21)
- DoctrineMongoODMModule: [requires `^3.4.0`](https://github.com/doctrine/DoctrineMongoODMModule/blob/83893c4bd76f70fe791f5beeec06d66d714c830d/composer.json#L31)

derrabus:
This would effectively mean that we don't maintain ORM 2 anymore, doesn't it?

greg0ire:
> From a [quick search](https://github.com/search?q=org%3Adoctrine+doctrine%2Fpersistence+language%3AJSON&type=code&l=JSON), I found the following packages that are not compatible with v4 yet


Good catch! I'll try to address that.

> This would effectively mean that we don't maintain ORM 2 anymore, doesn't it?

I think this is OK, because of this:

> Starting March 2026, we will only merge changes of the following kind into 2.x branches:

>    PHP version compatibility, let's say support for PHP 8.5, 8.6 or 9.0
>    Security fixes
>    Improving forward compatibility with ORM 3


https://www.doctrine-project.org/2025/10/08/an-update-on-the-orm-2-end-of-life.html

I propose we mark persistence 3 as unmaintained, but keep the branch around until ORM 2 is not  supported at all, in the unlikely case an extra patch release is needed.

GromNaN:
> I propose we mark persistence 3 as unmaintained, but keep the branch around until ORM 2 is not supported at all, in the unlikely case an extra patch release is needed.


"Keeping the branch to apply patches if needed" is an appropriate definition of maintaining the version. Therefore, to be clear in our communication, we should **not** mark this branch as unmaintained.

derrabus:
> "Keeping the branch to apply patches if needed" is an appropriate definition of maintaining the version. Therefore, to be clear in our communication, we should **not** mark this branch as unmaintained.

This!

greg0ire:
Ok… let's wait for the actual EOL then.

PR #521, opened 2026-09-03
Handle root-namespace classes in SymfonyFileLocator

### LINKED ISSUE #520: SymfonyFileLocator triggers an assertion for DateTime params
### Bug Report

| Version                                     | 3.6.8

#### Summary

Using DateTime (or other global namespace classes) with assertions enabled can cause an assertion failure in SymfonyFileLocator.

#### Current behavior

`assert(is_int($pos));` fails if $className doesn't contain a `\`, which is the case for DateTime.

#### Expected behavior

Not triggering an assert, obviously, but I'm not well-versed enough in doctrine internals to say how it should work. Either the assert should be removed (and the following exception adjusted accordingly) or `findMappingFile` shouldn't be called in the first place.

#### How to reproduce
A quite tiny repro:
```php
$locator = new Doctrine\Persistence\Mapping\Driver\SymfonyFileLocator([]);
$locator->findMappingFile(DateTime::class);
```

A more realistic way to trigger this is to call `\Doctrine\ORM\AbstractQuery::setParameter` with a DateTime as value but WITHOUT explicitly declaring the type in the third (type) parameter:
`$someQuery->setParam('foo', new DateTime());`

Junaid-PK:
Fixes #520.

When a mapping lookup fails for a class in the root namespace, `strrpos()` returns `false`. The assertion then fails when assertions are enabled; when they are disabled, the subsequent `substr()` starts at offset 1 and reports a filename such as `ateTime.yml`.

Use the complete class name when there is no namespace separator, while preserving the existing short-name behavior for namespaced classes. The regression test verifies that `DateTime` produces the expected `MappingException` and filename.

Validation:

- `vendor/bin/phpunit` — 146 tests, 306 assertions
- `vendor/bin/phpcs` — 72 files
- `vendor/bin/phpstan analyse --no-progress --memory-limit=512M` — no errors

### DIFF
--- a/src/Mapping/Driver/SymfonyFileLocator.php
+++ b/src/Mapping/Driver/SymfonyFileLocator.php
@@ -11,10 +11,8 @@ use RecursiveIteratorIterator;
 use RuntimeException;
 
 use function array_keys;
-use function assert;
 use function is_dir;
 use function is_file;
-use function is_int;
 use function realpath;
 use function sprintf;
 use function str_replace;
@@ -222,12 +220,12 @@ class SymfonyFileLocator implements FileLocator
             }
         }
 
-        $pos = strrpos($className, '\\');
-        assert(is_int($pos));
+        $pos            = strrpos($className, '\\');
+        $shortClassName = $pos === false ? $className : substr($className, $pos + 1);
 
         throw MappingException::mappingFileNotFound(
             $className,
-            substr($className, $pos + 1) . $this->fileExtension,
+            $shortClassName . $this->fileExtension,
         );
     }
 
--- a/src/Mapping/Driver/SymfonyFileLocator.php
+++ b/src/Mapping/Driver/SymfonyFileLocator.php
@@ -11,6 +11,7 @@ use RecursiveIteratorIterator;
 use RuntimeException;
 
 use function array_keys;
+use function basename;
 use function is_dir;
 use function is_file;
 use function realpath;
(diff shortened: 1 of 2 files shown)


greg0ire (dismissed):

src/Mapping/Driver/SymfonyFileLocator.php
```diff
@@ -222,12 +220,12 @@ public function findMappingFile(string $className): string
...
 
-        $pos = strrpos($className, '\\');
-        assert(is_int($pos));
+        $pos            = strrpos($className, '\\');
+        $shortClassName = $pos === false ? $className : substr($className, $pos + 1);
 
         throw MappingException::mappingFileNotFound(
             $className,
-            substr($className, $pos + 1) . $this->fileExtension,
+            $shortClassName . $this->fileExtension,
```
GromNaN: This also works:
```suggestion
throw MappingException::mappingFileNotFound(
$className,
            basename(str_replace('\\', '/', $className)) . $this->fileExtension,
```

GromNaN (dismissed):
Thanks for the fix!

stof (dismissed):

src/Mapping/Driver/SymfonyFileLocator.php
Junaid-PK: Good call, thanks. I applied this simplification in fb04176; the full test suite, PHPCS, and PHPStan still pass.

GromNaN (approved):

Issue #523, opened 2026-09-10
ObjectRepository::findBy orderBy should accept SortDirection


dmitryuk:
#### What
\Doctrine\Persistence\ObjectRepository::findBy
php doc for `$orderBy` argument should accept array<\SortDirection> enum
#### Why
Since php 8.6 (and symfony/polyfill-php86) provide \SortDirection enum

#### How
Fix phpdoc for \Doctrine\Persistence\ObjectRepository::findBy

Relates https://github.com/doctrine/orm/issues/11313

greg0ire:
On v4, it should accept it, and on v5, it must accept it, and no longer needs to accept strings.  Requiring that every implementation suddenly accept it would be a breaking change. Not sure what to do on 4.x. @derrabus @SenseException @stof @GromNaN thoughts? Maybe an `@method` annotation would help on 4.x? Also, requiring a polyfill just for a phpdoc comment feels weird.

GromNaN:
Thanks @dmitryuk for reporting this issue. The new enum will be added in 5.0.0 by https://github.com/doctrine/persistence/pull/525.

dmitryuk:
What about v4?

GromNaN:
See https://github.com/doctrine/persistence/pull/524#discussion_r3989565909
> Widening parameter type is a BC break for implementers (as implementers are not allowed to use a more restrictive type based on the Liskov Substitution Principle).

GromNaN:
Two PRs were opened to allow `SortDirection` values wherever sort directions are accepted:

- doctrine/orm#12613: `EntityRepository::findBy()` and `findOneBy()` now accept and document `SortDirection`, while strings remain non-deprecated, since the `ObjectRepository` interface of persistence 3.x and 4.x still documents strings as the only statically valid values.
- doctrine/mongodb-odm#3048: applies `SortDirection` to the document repository `findBy()`/`findOneBy()` and to the query builder, expression and aggregation sort orders.

PR #524, opened 2026-09-11
Allow SortDirection enum in ObjectRepository::findBy()

GromNaN:
Follows #523. Companion PR on 5.0.x: #525.

`ObjectRepository::findBy()` `$orderBy` now accepts the PHP 8.6 `SortDirection` enum in addition to the `'asc'`, `'desc'`, `'ASC'` and `'DESC'` string literals. The native signature is unchanged and strings remain accepted, so this is backward compatible for callers.

Implementations that do not already handle the enum must accept it and map it to their underlying sort direction, otherwise a caller passing `SortDirection` will trigger a runtime error.

Downstream support:

- `doctrine/orm` supports `SortDirection` in `findBy()` and `findOneBy()` starting with 3.7.0. Older versions (3.6.x and below, including the 2.20.x/2.21.x LTS lines) only accept strings and will not receive this change; upgrade to 3.7.0 or later to use the enum.
- `doctrine/mongodb-odm` supports `SortDirection` in `findBy()` starting with 2.18. Older versions (2.17.x and below) only accept strings and will not receive this change; upgrade to 2.18 or later. Its query builder `sort()` already accepts the enum since 2.17.0, but `findBy()` bypasses the query builder.

`symfony/polyfill-php86` is added to `require-dev` so PHPStan can resolve `SortDirection` on PHP 8.1-8.5 without a production dependency. The PHPDoc is validated by the static-analysis CI; the runtime test is a smoke test guarding against a future narrowing of the native type.

### DIFF
--- a/src/ObjectRepository.php
+++ b/src/ObjectRepository.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Persistence;
 
+use SortDirection;
 use UnexpectedValueException;
 
 /**
@@ -38,9 +39,9 @@ interface ObjectRepository
      * an UnexpectedValueException if certain values of the sorting or limiting details are
      * not supported.
      *
-     * @param array<string, mixed>       $criteria
-     * @param array<string, string>|null $orderBy
-     * @phpstan-param array<string, 'asc'|'desc'|'ASC'|'DESC'>|null $orderBy
+     * @param array<string, mixed>                     $criteria
+     * @param array<string, string|SortDirection>|null $orderBy
+     * @phpstan-param array<string, 'asc'|'desc'|'ASC'|'DESC'|SortDirection>|null $orderBy
      *
      * @return array<int, object> The objects.
      * @phpstan-return list<T>
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,34 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 4.3
+
+## `ObjectRepository::findBy()` `$orderBy` additionally accepts the `SortDirection` enum
+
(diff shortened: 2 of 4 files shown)


src/ObjectRepository.php
```diff
@@ -38,9 +39,9 @@ public function findAll(): array;
      * an UnexpectedValueException if certain values of the sorting or limiting details are
      * not supported.
      *
-     * @param array<string, mixed>       $criteria
-     * @param array<string, string>|null $orderBy
-     * @phpstan-param array<string, 'asc'|'desc'|'ASC'|'DESC'>|null $orderBy
+     * @param array<string, mixed>                     $criteria
+     * @param array<string, string|SortDirection>|null $orderBy
```
stof, line 43: Widening parameter type is a BC break for implementers (as implementers are not allowed to use a more restrictive type based on the Liskov Substitution Principle). So to me, this change must happen only in the 5.0.x branch.

GromNaN: You're absolutely right. And even if it doesn't break the PHP code because the type isn't enforced by PHP, we're providing information about a supported type that isn't respected by all older versions of ORM/ODM. Adding a composer conflict rule with older ORM versions would be really dirty.

The solution, therefore, is to accept `string|SortDirection` in 5.0 and (possibly) remove support for `string` in 6.0.

stof: > Adding a composer conflict rule with older ORM versions would be really dirty.

Indeed. An interface package that would be forced to add conflicts for older versions of implementation packages would be a clear signal that the interface package does not follow semver (assuming implementation packages don't use unbounded constraints of course but rely on semver constraints)

greg0ire: Can this maybe be achieved with `@method`?

stof: @greg0ire no. We don't want to add a new method. We want to change the signature of an existing method.

greg0ire: Ok, too bad there isn't an annotation to suggest this.

